PR #3, opened 2024-06-28
ingress implementations using k8s gateway api

IbraAoad:
First working version of the istio ingress charm

- This charm implements a holistic approach of managing the k8s ingress resources by making use of the [lightkube_extensions](https://github.com/canonical/lightkube-extensions/pull/2) python lib
- This charm will make use of the Traefik's v2 `ingress_per_app` interface and will not be compatible with v1 nor will it support `ingress_per_unit`

Some tasks are still a WIP and might be moved over to a different issue:

- Add proper mocks for scenario & unit tests and subsequently some tests
- Increase test coverage for itests
- Add support for defining external host instead of an IP for the ingress load balancer #6


src/charm.py
```diff
@@ -5,38 +5,388 @@
...
+class IstioIngressCharm(CharmBase):
     """Charm the service."""
 
     def __init__(self, *args):
         super().__init__(*args)
-
+        self.ingress_per_appv1 = IPAv1(charm=self)
         self.ingress_per_appv2 = IPAv2(charm=self)
+        self.lk_client = Client(namespace=self.model.name, field_manager="lightkube")
+        load_in_cluster_generic_resources(self.lk_client)
```
ca-scribner, line 58: Including this here means all unit/scenario tests will have to mock out the client, and mocking the client embedded in here is a pain.  What about moving the lk client into a helper method?  It can do the `load_in_cluster_generic_resources` and return a client instance.  That'll reduce when you need a functioning client mock to just when you're actually using it.

src/charm.py
```diff
@@ -5,38 +5,388 @@
 
 """Istio Ingress Charm."""
 
+import functools
 import logging
+import time
+from typing import Any, Dict, Optional, Union
 
-import ops
+from charms.traefik_k8s.v1.ingress import IngressPerAppProvider as IPAv1
```
ca-scribner, line 13: Do we need to provide for both versions of this interface?  It'll be a while before anyone uses this charm, and they'll be refactoring their app to use it anyway, so maybe we can also require them to use the most recent version

src/charm.py
```diff
@@ -5,38 +5,388 @@
...
+
+        # Set status to Active
+        self.unit.status = ActiveStatus(f"Serving at {self._external_host}")
+
+    def _handle_ingress_data_provided(self, event: RelationEvent):
         """Handle a unit providing data requesting IPU."""
-        self.unit.status = ops.BlockedStatus("Not yet implemented")
+        self.unit.status = MaintenanceStatus("Setting up an ingress")
+        self._process_ingress_relation(event.relation)
+        self.unit.status = ActiveStatus(f"Serving at {self._external_host}")
```
ca-scribner, line 165: Have you considered taking a holistic approach to reconciling the charm (rebuilding the whole world on all changes), rather than an imperative approach (changing things based on this specific event's needs)?  I think the holistic approach will save you a lot of headaches later.  

One example where imperative is going to be tough:
* user establishes two relations (to App1 and App2)
* App1 provides **invalid** data.  App1's relation_changed event fires first and  `_push_configurations` breaks, putting the charm into `BlockedStatus`
* App2 provides **valid** data.  App2's relation_changed event fires, `_push_configurations` works, and the charm goes into `Active` even though App1's ingress is not satisfied

Contrasting to the holistic approach, on every change event you can loop through all relations and construct HTTPRoutes accordingly.  The charm always has full information about its desired state, so just recompute everything.  That way when serving App2's HTTPRoute, you still know App1's is broken and can handle that accordingly.  An implementation of this is [here in istio-pilot](https://github.com/canonical/istio-operators/blob/98d6b2f58f7c6da5c5a4a319f1a7ce48ce671874/charms/istio-pilot/src/charm.py#L671), which accepted `routes` (data from all ingress relations), rendered that into a manifest with all `VirtualServices`, and `reconciled` that manifest against what is currently in cluster.  

An alternative approach might be keeping information like this in storedstate, but it is fragile.  Every event needs to be handled perfectly otherwise you'll have a hidden error, and its hard to predict every possible event sequence.

src/charm.py
```diff
@@ -5,38 +5,323 @@
...
+        for route in routes_to_delete:
+            self.crd_manager.delete_resource(
+                name=route["name"],
+                namespace=route["namespace"],
+                resource_type=route["resource_type"],
+            )
+        for rel, ingress_data in current_ingresses:
+            self._create_ingress(rel, ingress_data)
+
+    def _create_ingress(self, rel, ingress_data):
```
ca-scribner, line 227: This is more a preference than something with a right or wrong answer, but I think a jinja template for this yaml would be more descriptive than building this typed resource.  Main reason being that someone familiar with an HTTPRoute can look at a jinja template and it looks familiar.  Might not be fully fleshed out, but it is close enough to know what the end product will include

src/charm.py
```diff
@@ -5,38 +5,323 @@
...
+                "name": data.app.name,
+                "namespace": data.app.model,
+                "resource_type": HTTPROUTE_RESOURCE,
+                "gateway_name": self.app.name,
+                "gateway_model_name": self.model.name,
+                "prefix": prefix,
+                "strip_prefix": data.app.strip_prefix,
+                "backend_svc": data.app.name,
+                "backend_port": data.app.port,
+            }
```
ca-scribner, line 208: Can't we construct an httproute resource here, rather than a dict with the information? I see this gets passed to `_create_ingress`, but then why not do that right away?

src/charm.py
ca-scribner, line 227: this is effectively `def _create_ingress(self, *args, more_args_in_a_dict)` - shouldn't the content of `ingress_data` just be keyword args in the signature of `_create_ingress`?

src/charm.py
```diff
@@ -5,38 +5,323 @@
...
+        for rel in self.model.relations["ingress"]:
+
+            if not self.ingress_per_appv2.is_ready(rel):
+                logger.debug(f"Provider {rel} not ready; resetting ingress configurations.")
+                self.ingress_per_appv2.wipe_ingress_data(rel)
+                continue
+
+            if not rel.app:
+                logger.error(f"No app on relation {rel}, Skipping ingress configuration.")
+                continue
```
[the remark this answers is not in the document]
IbraAoad: It seems unlikely I'd say but I'm not sure, I duplicated most of the guards implemented in [traefik](https://github.com/canonical/traefik-k8s-operator/blob/90b209c7435bf29600fcaed52be4d79c2816fc8f/src/charm.py#L923), I removed it for now and we can add it back if we observed smth similar

IbraAoad:
Integration tests are failing because [istio-k8s-operator](https://github.com/canonical/istio-k8s-operator) is a dependency for the ingress provider to work, specifically the `gateway.networking.k8s.io` CRDs.

src/charm.py
```diff
@@ -5,38 +5,344 @@
...
+            if not self.ingress_per_appv2.is_ready(rel):
+                logger.debug(f"Provider {rel} not ready; resetting ingress configurations.")
+                self.ingress_per_appv2.wipe_ingress_data(rel)
+                continue
+
+            try:
+                data = self.ingress_per_appv2.get_data(rel)
+            except DataValidationError as e:
+                logger.error(f"Data validation error for relation {rel}: {e}.")
+                raise e
```
ca-scribner, line 275: Maybe still good to leave this here, but it isn't possible to get here if we guard against `is_ready()==False` above.  `is_ready()` calls `get_data`

Can we combine the above `is_ready` and this `get_data` to just a single `try`/`except` around a `get_data`?

src/charm.py
```diff
@@ -5,38 +5,344 @@
...
+
+            prefix = self._generate_prefix(data.app.dict(by_alias=True))
```
ca-scribner, line 277: I think this is a limitation in the ingress interface/library, but `prefix` should be configurable by the related application.  These define the structure of someone's web application.  If this is a limit in the interface/library, we should open an issue here so we can extend the interface

src/charm.py
```diff
@@ -5,38 +5,344 @@
...
+            resource_to_append = self._construct_httproute(data, prefix)
```
ca-scribner, line 278: I think this is a question more about the ingress interface/library, but ...

Is a related application limited to a single `HTTPRoute` with a single `rule`?  That sounds restrictive, but I can't think of a use case that breaks this assumption.  Maybe a charm that has multiple workloads, but I can't think of one.

src/charm.py
```diff
@@ -5,38 +5,344 @@
...
+                    Rule(
+                        matches=[Match(path=PathMatch(value=prefix))],
+                        backendRefs=[
+                            BackendRef(
+                                name=data.app.name,
+                                port=data.app.port,
+                                namespace=data.app.model,
+                            )
+                        ],
+                        filters=([URLRewriteFilter()] if data.app.strip_prefix else []),
```
ca-scribner, line 247: I think this is a limitation of the interface/library, but...

Do you know how `strip_prefix` functions?  Like given a user hitting `http://INGRESS_URL/PREFIX`, what URL reaches the application if `strip_prefix=True` vs `=False`?  My guess is:
* False: `http://INGRESS_URL/PREFIX`
* True: one of
  * `/PREFIX`
  * `/`

?

This might be a traefik vs istio difference, but I'd have expected a `rewrite` arg here that lets the requesting app say exactly what this url look like

src/charm.py
```diff
@@ -5,38 +5,344 @@
...
+        self.unit.status = MaintenanceStatus("Setting up global istio-ingress gateway")
+        resource_to_append = self._construct_gateway()
+        resources_list.append(resource_to_append)
+        krm.reconcile(resources_list)
 
-    def _handle_ingress_data_provided(self, _):
+        self.unit.status = MaintenanceStatus("Validating gateway readiness")
+
+        if self._is_ready():
+            self.unit.status = ActiveStatus(f"Serving at {self._external_host}")
```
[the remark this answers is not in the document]
IbraAoad: I don't think `on_collect_status` is needed since it's now all combined into 1 main reconciler

src/charm.py
```diff
@@ -5,38 +5,344 @@
...
+        return None
+
+    if not (status := service.status):
+        return None
+    if not (load_balancer_status := status.loadBalancer):
+        return None
+    if not (ingress_addresses := load_balancer_status.ingress):
+        return None
+    if not (ingress_address := ingress_addresses[0]):
+        return None
```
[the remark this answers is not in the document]
IbraAoad: We don't really need to handle cases for `nodeport` or `clusterIP` since we're explicitly setting the gateway to use the a default k8s lb, also based on our design decisions of having 1 main `gateway` I'd say an lb here works best, for cases when len(ingresses)>1 we have a couple of scenarios

1. multiple ips with 1 hostname (Here we don't need to do changes, using the hostname will allow k8s to do it's own load-balancing across the ips)
2. multiple ips with multiple hostnames (I'd say this is an extremely rare case that will require us to figure out a way to balance among them before going through k8s balancing, I'd say this can be deferred for now until we actually have a need for it)

src/charm.py
```diff
@@ -5,38 +5,344 @@
...
+        resources_list = []
+        krm = self._get_gateway_resource_manager()
 
-    def _on_start(self, _):
-        """Event handler for start."""
-        self.unit.status = ops.ActiveStatus()
+        self.unit.status = MaintenanceStatus("Setting up global istio-ingress gateway")
+        resource_to_append = self._construct_gateway()
+        resources_list.append(resource_to_append)
+        krm.reconcile(resources_list)
```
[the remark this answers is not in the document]
IbraAoad: was initially like your suggestion, but I had to make this change to "please" the static checker, I can always change it back and add a pyright skip if that seems more reasonable.

src/charm.py
IbraAoad: to be tracked in #4

src/models.py
```diff
@@ -0,0 +1,120 @@
...
+    port: int = 80
+    protocol: str = "HTTP"
+    allowedRoutes: AllowedRoutes = AllowedRoutes()  # noqa: N815
+
+
+class GatewaySpec(BaseModel):
+    """GatewaySpec defines the specification of a gateway."""
+
+    gatewayClassName: str = "istio"  # noqa: N815
+    listeners: List[Listener] = Field(default_factory=lambda: [Listener()])
```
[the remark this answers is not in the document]
IbraAoad: to be tracked in #6

src/models.py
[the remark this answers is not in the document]
dstathis: While I agree that the jinja *looks* much better, I think we can handle understanding it when written in python. The benefit of writing it in python is that it is much easier to manipulate the objects as our use cases get more and more complex. So my vote is for pydantic models.

requirements.txt
```diff
@@ -1,2 +1,10 @@
 ops ~= 2.5
 pydantic>=2
+
+lightkube-extensions @ git+https://github.com/canonical/lightkube-extensions.git@add-initial-helpers
```
dstathis, line 4: Whenever you put a git link as a python requirement, you should have a comment explaining why so that in the future we should know if it is still correct.

src/charm.py
```diff
@@ -6,37 +6,365 @@
...
-        """Event handler for start."""
-        self.unit.status = ops.ActiveStatus()
+    def _on_remove(self, _):
+        """Event handler for remove."""
+        # Removing tailing ingresses
+        kim = self._get_ingress_resource_manager()
+        kim.delete()
 
-    def _handle_ingress_data_provided(self, _):
+        kgm = self._get_gateway_resource_manager()
```
dstathis, line 143: Maybe I'm missing something obvious but what does "kgm" mean?

src/charm.py
```diff
@@ -6,37 +6,365 @@
...
+                    deployment.status
+                    and deployment.status.readyReplicas == deployment.status.replicas
+                ):
+                    return True
+            except ApiError as e:
+                logger.error(f"Error checking gateway deployment status: {e}")
+
+            time.sleep(10)
+
+        return False
```
dstathis, line 173: I feel like there will likely be clouds where 100 seconds is always not enough time. Maybe we can make this tune-able with a config variable?

tox.ini
```diff
@@ -59,14 +70,14 @@ deps =
...
-                 {[vars]tests_path}/scenario
-    coverage report
+    ; coverage run --source={[vars]src_path} \
+    ;              -m pytest \
+    ;              --tb native \
+    ;              -v \
+    ;              -s \
+    ;              {posargs} \
+    ;              {[vars]tests_path}/scenario
+    ; coverage report
```
dstathis, line 80: Either delete it or don't.

dstathis (approved):

src/charm.py
IbraAoad: convention is just based on the [lightkube-extensions](https://github.com/canonical/lightkube-extensions/blob/5469f00f8b1333fc1e6861899d18c9418de77cab/src/lightkube_extensions/batch/_kubernetes_resource_manager.py#L79) lib where `krm` would be `k8s_resource_manager` and `kgm`: `k8s_gateway_manager`, but now that I'm re-reading this I feel it's very ambiguous, so I switched all references of this in the code to be `krm`.

PR #10, opened 2024-08-09
adding hostname support

### LINKED ISSUE #6: Add support for defining external host instead of an IP for the ingress load balancer
### Enhancement Proposal

a `Gateway` resource supports defining a [hostname](https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io/v1.Listener) to be used as the ingress lb address, we need to add support to the current istio ingress implementation to start pulling that from [requirer relation data](https://github.com/canonical/traefik-k8s-operator/blob/f131f2a4c7997e329c402170fc43bfa5ec17de7c/lib/charms/traefik_k8s/v2/ingress.py#L630)

IbraAoad:
## Issue
Fixes #6 

## Solution
Adding support in istio ingress to allow user to pass an external hostname via charm config to be used in the gateway resource


src/models.py
```diff
@@ -29,6 +29,8 @@ class AllowedRoutes(BaseModel):
 class Listener(BaseModel):
     """Listener defines a port and protocol configuration."""
 
+    model_config = ConfigDict(extra="allow")
```
ca-scribner, line 32: Why do this instead of add the optional hostname to the listener?

src/charm.py
```diff
@@ -369,6 +378,54 @@ def _generate_prefix(data: Dict[str, Any]) -> str:
         name = data["name"].replace("/", "-")
         return f"/{data['model']}-{name}"
 
+    def _is_valid_hostname(self, hostname: Optional[str]) -> bool:
```
ca-scribner, line 381: Can we add some unit tests for this?

ca-scribner: Are there any standard packages that have this check we can consume?  This feels like something commonly done.

charmcraft.yaml
ca-scribner, line 1: We need an integration test on this to demonstrate we know how this works (and that we know it works with metallb).  iiuc, this keys off a HOST header like described [here](https://stackoverflow.com/a/43156094), so I think we could:
* deploy microk8s with metallb
* deploy istio-ingress with hostname=some.host.name.  This'll create an ingress with the standard metallb external IP (10.whatever.from.metallb)
* relate a test workload to get an httproute.  We need some sort of hello world workload that'll respond back (there's some in istio upstream if you want those)
* `curl 10.whatever.from.metallb/test-workload-route`, which should do nothing (because our curl will not have the correct host)
* `curl 10.whatever.from.metallb/test-workload-route -H Host=some.host.name` (I think my syntax here is wrong, but close enough), which should successfully hit our demo workload

Rather than curl, probably makes more sense to use requests, but you get the idea.

ca-scribner (commented):
a few comments but generally looks good!

src/models.py
IbraAoad: This was to prevent pydantic from passing it to the kube api as None, but digging into the pydantic docs it seems you can do `exclude_none=True` on data serialization so it should now be fixed, Thaaaaanks 😊

src/charm.py
IbraAoad: Unit tests are added, for the std package I'd rather use the same regex as the [upstream project](https://github.com/kubernetes-sigs/gateway-api/blob/6446fac9325dbb570675f7b85d58727096bf60a6/apis/v1/shared_types.go#L523) tbh as their implementation of a hostname isn't the industry standard, this should be documented well in the code tho

charmcraft.yaml
IbraAoad: itests are added to validate 
- charm is reachable with the MetalLB IP
- gateway is up, healthy and has the correct host
- HTTPRoute is validated and accepted with no errors and
- Route is attached correctly to the gateway listeners


However doing, `curl 10.whatever.from.metallb/test-workload-route -H Host=some.host.name` and `curl 10.whatever.from.metallb/test-workload-route` will both work for now as I've omitted the usage of hostnames in the route and went for the static global host in the gateway so curling the metalLB IP with and without a host header will reach the same global listener, I've also left a few TODO comments in the code to explain this and what to do on future dev iterations.

tests/integration/test_charm_ipa.py
ca-scribner, line 1: I think we should refactor `test_ipa_charm_has_ingress`.  afaict, the current version can be a unit test (it relates charms then asserts the relation data is as expected).  

What would be helpful to integration test here though is whether the route actually works.  For example, we dont do an equivalent to `curl URL`., thus we don't know if our Gateway and HTTPRoute are configured correctly.  imo that's what we should test here

I propose we:
* add a workload to the tester charm.  I like [echo-server](https://github.com/jmalloc/echo-server) for stuff like this because it replies back everything about the request its sent.  See below for example deployment configuration I use in debugging
* add real access attempts on the ingress URL to prove the tester app's workload is reachable.  We can do this both with the IP as the LB host, as well as if we set host to foo.bar

Echo server:
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: echo-deployment
spec:
  replicas: 1
  selector:
    matchLabels:
      app: echo-server
  template:
    metadata:
      labels:
        app: echo-server
    spec:
      containers:
        - name: echo-server
          image: jmalloc/echo-server
          ports:
            - name: http-port
              containerPort: 8080
          env:
            - name: LOG_HTTP_HEADERS
              value: STDOUT
            - name: LOG_HTTP_BODY
              value: STDOUT
```

src/charm.py
ca-scribner, line 1: afaict, changing the hostname back from a hostname does not work as expected.  for example:

```bash
juju deploy istio-ingress-k8s
# gateway has host=IP
juju config istio-ingress-k8s external_hostname=foo.bar
# gateway has host=foo.bar
juju config istio-ingress-k8s external_hostname=""
# (incorrect) gateway has host=foo.bar
juju config istio-ingress-k8s external_hostname=10.64.140.43
# (incorrect) gateway has host=foo.bar
# debug-logs show the charm parsing 10.64.140.43 as an invalid hostname (although it is still a valid value in gateway's host field)
```

This would be a good thing to check in unit tests

ca-scribner (changes requested):

src/charm.py
```diff
@@ -18,6 +19,16 @@
...
 from lightkube_extensions.batch import KubernetesResourceManager, create_charm_default_labels
+from ops import BlockedStatus
+from ops.charm import (
+    CharmBase,
+)
+from ops.main import main
+from ops.model import (
+    ActiveStatus,
+    MaintenanceStatus,
+)
```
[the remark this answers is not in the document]
IbraAoad: not sure tbh, but it seems the linter here accepts both as valid syntax

tests/integration/helpers.py
```diff
@@ -30,6 +31,106 @@ async def get_k8s_service_address(ops_test: OpsTest, service_name: str) -> Optio
...
+    Args:
+        ops_test: pytest-operator plugin
+        gateway_name: Name of the Gateway resource.
+
+    Returns:
+        A dictionary representing the status of the first listener, or None if not found.
+    """
+    model = ops_test.model.info
+    try:
+        result = sh.kubectl(
```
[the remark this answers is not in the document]
IbraAoad: No really a reason in particular, just seemed to be easier and more convenient in usage for testing compared to lightkube, especially for CRDs like gateways and routes.

src/charm.py
IbraAoad: [IPs shouldn't be valid](https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io%2fv1.Hostname) and they get rejected on purpose, the incorrect hostname after setting the config to an empty string is most probably due to the patch strategy of the reconciler, I'm still investigating tho.

ca-scribner: Well this is an interesting, because I'm testing it right now and it works.  Try it out by changing the gateway host to `10.64.140.43` and then doing `curl -v 10.64.140.43/... -H 'Host: 10.64.140.43'`

I do see where it says specifically it wont work though.  So I'm not sure what to make of it.  I think we have to support this though, otherwise metallb just doesnt work at all as a LB for us does it?

ca-scribner: sorry, I should clarify that hostnames as IP addresses work.

But changing the external_hostname="" does not unset the hostname on the gateway

ca-scribner: I think I'm seeing the same thing you mentioned today in the daily.  I thought the issue was that we needed to set `hostname=None`.  There's discussion [here](https://github.com/gtsystem/lightkube/issues/43).  

I tried changing the construct_gateway to use:
```python
        hostname = self._external_host if self._is_valid_hostname(self._external_host) else None

        gateway = IstioGatewayResource(
            metadata=Metadata(
                name=self.app.name,
                namespace=self.model.name,
            ),
            spec=IstioGatewaySpec(
                gatewayClassName="istio",
                listeners=[
                    Listener(
                        name="default",
                        port=80,
                        protocol="HTTP",
                        allowedRoutes=AllowedRoutes(namespaces={"from": "All"}),
                        hostname=hostname,
                    )
                ],
            ),
        )
```
where we set hostname=None if its unset.  It didn't quite work how I hoped, but I ran out of time to figure out why.  Might have been because of exclude_none=True.

I did find that, if I manually do a `.patch` using this object instead of `.apply`, PatchType.MERGE seemed to work better.  Was that what you were mentioning today?  But maybe we can get the previous one to work too, im not sure

tests/integration/helpers.py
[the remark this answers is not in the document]
IbraAoad: afaik it's installed by default with the [actions operator](https://github.com/charmed-kubernetes/actions-operator/blob/5193238c86d58e874ad3457da2df640ef138a41f/src/bootstrap/index.ts#L376), yet I'm not really opinionated against using lk, both are fine by me :D

src/charm.py
IbraAoad: For the IPs, I'd rather stick to the gateway guides and assume that it's a bug in their validation webhooks, tbh it makes sense not to allow IPs as I don't see a use case where you'd want to set an IP (not hostname) instead of the one already generated from your LB.

As for the whole patch strategy, I've introduce sort of a hack that fixes this and it should be well documented as well

ca-scribner: I think IP-as-hostname (or at least IP.nip.io) will be a hard requirement on us because of Kubeflow local installs, but we don't need to block this PR on that discussion.  I've opened #13 to handle that

ca-scribner: For the patch strategy, this fix effectively turns the gateway (briefly) into a wildcard gateway on every charm reconcile.  That feels like we're asking for trouble.  Probably fine for a POC, but not for real users (would it have security concerns?  stability?  not sure..)

what about refactoring to:
```
    def _sync_gateway_resources(self):
        resources_list = []
        krm = self._get_gateway_resource_manager()

        resource_to_append = self._construct_gateway()
        resources_list.append(resource_to_append)
        krm.reconcile(resources_list)

        if not self._is_valid_hostname(self._external_host):
            self._remove_hostname_if_present()
```

That way it only removes the hostname when it should be removed.

I'm also not sure if `remove_hostname_if_present` should remove the hostname on just `listener[0]` or on all listeners - can you explain the reasoning behind listener[0]?  Is that because we're only ever deploying one Listener?  If yes, we should add a note in the `construct_gateway` Listener definition to say this other part of the code prevents using multiple Listeners, that way we don't forget later.

src/charm.py
```diff
@@ -361,14 +385,55 @@ def _external_host(self) -> Optional[str]:
         If the gateway isn't available or doesn't have a load balancer address yet,
         returns None. Only use this directly when external_host is allowed to be None.
         """
-        return self._get_lb_status
+        if external_hostname := self.model.config.get("external_hostname"):
+            hostname = cast(str, external_hostname)
+            if self._is_valid_hostname(hostname):
+                return hostname
+            logger.warning("Invalid hostname provided, defaulting to loadbalancer external IP")
```
ca-scribner, line 392: Sorry I didn't catch this before.  

I don't think we should automatically resolve this.  I don't see why `juju config istio-ingress-k8s hostname="something-incorrect"` can be interpreted as `hostname=loadbalancerIP`.  I'd expect to see an error or blocked status somewhere more obvious (juju status), and definitely not Active status.  I don't expect anyone seeing Active will go looking through debug-log anyway just in case

src/charm.py
IbraAoad: Reason was just that we only now have 1 listener which is obviously going to change when introducing TLS, anyways I've updated it now to remove hosts for all listeners.

tests/integration/test_charm_ipa.py
```diff
@@ -77,21 +104,64 @@ async def test_ipa_charm_has_ingress(ops_test: OpsTest):
...
+    await ops_test.model.wait_for_idle([APP_NAME, "ipa-tester"])
+
+    data = get_relation_data(
+        requirer_endpoint="ipa-tester/0:ingress",
+        provider_endpoint="istio-ingress-k8s/0:ingress",
+        model=ops_test.model_full_name,
+    )
+
+    requirer_app_data = data.requirer.application_data
+    model = dequote(requirer_app_data["model"])
```
ca-scribner, line 137: I think this simplification works as well:
```suggestion
    model = ops_test.model.name
```

I've pushed a commit with this change, but feel free to change it back if I missed something

ca-scribner (approved):
lgtm!  Great job

I pushed 2 small changes during my review, so check them and make sure you agree before you merge.  If you disagree, please change!

PR #15, opened 2024-10-03
feat: add TLS support for the gateway

ca-scribner:
## Issue
Implementing TLS for the ingress using the tls-certificate-interface (OPENG-2525)

## Solution
This implements TLS support using the tls-certificates-interface.  If related to a certificate provider, the gateway produced by the ingress will require TLS communication.

Design decisions (subject to discussion - could be changed):
* when TLS is successfully enabled, only an HTTPS listener is provided (the HTTP listener is removed)

## Testing Instructions

(I had to write these by memory at the end of the day - sorry if there's a typo.  feel free to edit!)

```
# Dependencies
juju add-model istio-system
juju deploy istio-k8s -m istio-system --channel edge --trust
juju deploy self-signed-certificates

# Tester charm (something that needs an ingress)
cd tests/integration/testers/ipa
charmcraft pack
juju deploy ./ipa-requirer-mock_ubuntu-20.04-amd64.charm --resource echo-server-image=jmalloc/echo-server:v0.3.7 --trust ipa-tester

# ingress
charmcraft pack 
juju deploy ./istio-ingress-k8s_ubuntu-22.04-amd64.charm --trust
juju relate istio-ingress-k8s:ingress ipa-tester:ingress

# demonstrate ipa-requirer-mock is reachable via the ingress
curl -v  http://10.64.140.43/istio-system-ipa-tester
# gives 200

# enable TLS
juju relate istio-ingress-k8s self-signed-certificates

# demonstrate ipa-tester is reachable via https (using -k, which ignores cert verification)
curl -v -k https://10.64.140.43/istio-system-ipa-tester

# and to confirm the certs, get the CA cert from self-signed-certificates
juju run self-signed-certificates/leader get-ca-certificate --output ca-certificate.temp
cat ca-certificate.temp | yq .ca-certificate > ca-certificate.pem 
rm ca-certificate.temp
# cert is available in ca-certificate.pem

curl -v --cacert ca-certificate.pem https://10.64.140.43/istio-system-ipa-tester

# confirm if we set a hostname, everything still works
juju config istio-ingress-k8s external_hostname="foo.bar"
# wait for everything to settle

# curl, but include a Host header and resolve foo.bar to the ingress IP
# if we don't do these things, the certificate wont verify successfully
curl -v \
 -HHost:foo.bar \
 --resolve "foo.bar:443:10.64.140.43" \
 --cacert ca-certificate.pem \
 "https://foo.bar:443/istio-system-ipa-tester"
# gives 200
```

## Todo:
* [x] integration test.  The feature is manually tested and previously worked, but integration tests aren't working.
* [x] if we have a TLS relation but no certs yet, should http still be available?


src/charm.py
```diff
@@ -96,6 +106,25 @@ def __init__(self, *args):
             self.ingress_per_appv2.on.data_removed, self._on_ingress_data_removed
         )
 
+        # During the initialisation of the charm, we do not have a LoadBalancer and thus a LoadBalancer external IP.
+        # If we need that IP to request the certs, disable cert handling until we have it.
```
ca-scribner, line 110: Interested in feedback on this approach.  The issue is that since the charm as a backup use the LoadBalancer IP as the hostname, there's situations where we wont have that IP yet (for example, before we've created the gateway) and thus we cannot properly configure the CertHandler.  An option to avoid this would be to never use the LoadBalancer IP for our gateway host (maybe in that case just leave a wildcard gateway host?), and then make a wildcard cert

For the current approach, I worry that we could fully instantiate the charm and never see a LoadBalancer IP (ex: if the LoadBalancer provider is slow or isn't set up yet) which I think will leave our charm stuck never setting up the TLS configuration

src/charm.py
```diff
@@ -220,28 +253,61 @@ def _is_ready(self) -> bool:
...
+        if tls_secret_name:
+            tls_configuration = GatewayTLSConfig(
+                certificateRefs=[SecretObjectReference(name=tls_secret_name)]
+            )
+            protocol = "HTTPS"
+            port = 443
+        else:
+            tls_configuration = None
+            protocol = "HTTP"
+            port = 80
```
ca-scribner, line 285: When TLS is enabled, should we do:
* disable HTTP, enable HTTPS
* enable HTTP and HTTPS

?

dstathis: I would say disable http for sure

IbraAoad: I'd like to suggest an alternative,

Keep the Gateway configs the same and implement [http-to-https-redirects](https://gateway-api.sigs.k8s.io/guides/http-redirect-rewrite/#http-to-https-redirects)

ca-scribner: I like @IbraAoad's suggestion.  I'll implement that and see how it looks

ca-scribner: Implemented.  let me know what you think!

src/charm.py
IbraAoad: Isn't this an acceptable behaviour I guess?, the is_ready guard is setting the charm in blocked state if an ip isn't given by the LB and we won't be handing out routes (neither http nor https will be working), I think main issue here is what will wake up the charm if the LB is finally ready to hand out an IP, having a k8s watcher would be the way to do it but still feels incorrect to put inside the charm.

I'm thinking we can have a seperate issue for this to investigate a solution for both this charm and for traefik cause it kinda have the same issue.

src/charm.py
```diff
@@ -354,6 +480,10 @@ def _sync_ingress_resources(self):
             raise RuntimeError("Ingress can only be provided on the leader unit.")
 
         krm = self._get_ingress_resource_manager()
+
```
IbraAoad, line 483: I get the below error when testing 

```bash
unit-istio-ingress-k8s-0: 12:07:44 ERROR unit.istio-ingress-k8s/0.juju-log certificates:16: Uncaught exception while in charm code:
Traceback (most recent call last):
  File "/var/lib/juju/agents/unit-istio-ingress-k8s-0/charm/venv/lightkube/core/generic_client.py", line 232, in raise_for_status
    resp.raise_for_status()
  File "/var/lib/juju/agents/unit-istio-ingress-k8s-0/charm/venv/httpx/_models.py", line 763, in raise_for_status
    raise HTTPStatusError(message, request=request, response=self)
httpx.HTTPStatusError: Client error '422 Unprocessable Entity' for url 'https://10.152.183.1/apis/gateway.networking.k8s.io/v1/namespaces/istio-system/httproutes/ipa-requirer-mock-http?fieldManager=istio-ingress-k8s'
For more information check: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/422

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/var/lib/juju/agents/unit-istio-ingress-k8s-0/charm/./src/charm.py", line 615, in <module>
    main(IstioIngressCharm)
  File "/var/lib/juju/agents/unit-istio-ingress-k8s-0/charm/venv/ops/main.py", line 45, in main
    return _main.main(charm_class=charm_class, use_juju_for_storage=use_juju_for_storage)
  File "/var/lib/juju/agents/unit-istio-ingress-k8s-0/charm/venv/ops/_main.py", line 543, in main
    manager.run()
  File "/var/lib/juju/agents/unit-istio-ingress-k8s-0/charm/venv/ops/_main.py", line 529, in run
    self._emit()
  File "/var/lib/juju/agents/unit-istio-ingress-k8s-0/charm/venv/ops/_main.py", line 518, in _emit
    _emit_charm_event(self.charm, self.dispatcher.event_name, self._juju_context)
  File "/var/lib/juju/agents/unit-istio-ingress-k8s-0/charm/venv/ops/_main.py", line 134, in _emit_charm_event
    event_to_emit.emit(*args, **kwargs)
  File "/var/lib/juju/agents/unit-istio-ingress-k8s-0/charm/venv/ops/framework.py", line 347, in emit
    framework._emit(event)
  File "/var/lib/juju/agents/unit-istio-ingress-k8s-0/charm/venv/ops/framework.py", line 857, in _emit
    self._reemit(event_path)
  File "/var/lib/juju/agents/unit-istio-ingress-k8s-0/charm/venv/ops/framework.py", line 947, in _reemit
    custom_handler(event)
  File "/var/lib/juju/agents/unit-istio-ingress-k8s-0/charm/lib/charms/tls_certificates_interface/v3/tls_certificates.py", line 1924, in _on_relation_changed
    self.on.certificate_available.emit(
  File "/var/lib/juju/agents/unit-istio-ingress-k8s-0/charm/venv/ops/framework.py", line 347, in emit
    framework._emit(event)
  File "/var/lib/juju/agents/unit-istio-ingress-k8s-0/charm/venv/ops/framework.py", line 857, in _emit
    self._reemit(event_path)
  File "/var/lib/juju/agents/unit-istio-ingress-k8s-0/charm/venv/ops/framework.py", line 947, in _reemit
    custom_handler(event)
  File "/var/lib/juju/agents/unit-istio-ingress-k8s-0/charm/lib/charms/observability_libs/v1/cert_handler.py", line 528, in _on_certificate_available
    self.on.cert_changed.emit()  # pyright: ignore
  File "/var/lib/juju/agents/unit-istio-ingress-k8s-0/charm/venv/ops/framework.py", line 347, in emit
    framework._emit(event)
  File "/var/lib/juju/agents/unit-istio-ingress-k8s-0/charm/venv/ops/framework.py", line 857, in _emit
    self._reemit(event_path)
  File "/var/lib/juju/agents/unit-istio-ingress-k8s-0/charm/venv/ops/framework.py", line 947, in _reemit
    custom_handler(event)
  File "/var/lib/juju/agents/unit-istio-ingress-k8s-0/charm/./src/charm.py", line 162, in _on_cert_handler_cert_changed
    self._sync_all_resources()
  File "/var/lib/juju/agents/unit-istio-ingress-k8s-0/charm/./src/charm.py", line 424, in _sync_all_resources
    self._sync_ingress_resources()
  File "/var/lib/juju/agents/unit-istio-ingress-k8s-0/charm/./src/charm.py", line 516, in _sync_ingress_resources
    krm.reconcile(current_ingresses)
  File "/var/lib/juju/agents/unit-istio-ingress-k8s-0/charm/venv/lightkube_extensions/batch/_kubernetes_resource_manager.py", line 179, in reconcile
    self.apply(resources=resources, force=force)
  File "/var/lib/juju/agents/unit-istio-ingress-k8s-0/charm/venv/lightkube_extensions/batch/_kubernetes_resource_manager.py", line 102, in apply
    apply_many(
  File "/var/lib/juju/agents/unit-istio-ingress-k8s-0/charm/venv/lightkube_extensions/batch/_many.py", line 70, in apply_many
    returns[i] = client.apply(
  File "/var/lib/juju/agents/unit-istio-ingress-k8s-0/charm/venv/lightkube/core/client.py", line 754, in apply
    return self.patch(
  File "/var/lib/juju/agents/unit-istio-ingress-k8s-0/charm/venv/lightkube/core/client.py", line 489, in patch
    return self._client.request(
  File "/var/lib/juju/agents/unit-istio-ingress-k8s-0/charm/venv/lightkube/core/generic_client.py", line 308, in request
    return self.handle_response(method, resp, br)
  File "/var/lib/juju/agents/unit-istio-ingress-k8s-0/charm/venv/lightkube/core/generic_client.py", line 248, in handle_response
    self.raise_for_status(resp)
  File "/var/lib/juju/agents/unit-istio-ingress-k8s-0/charm/venv/lightkube/core/generic_client.py", line 234, in raise_for_status
    raise transform_exception(e)
lightkube.core.exceptions.ApiError: HTTPRoute.gateway.networking.k8s.io "ipa-requirer-mock-http" is invalid: [spec.rules[0].filters[0].urlRewrite: Invalid value: "null": spec.rules[0].filters[0].urlRewrite in body must be of type object: "null", <nil>: Invalid value: "null": some validation rules were not checked because the object was invalid; correct the existing errors to complete validation]
unit-istio-ingress-k8s-0: 12:07:44 ERROR juju.worker.uniter.operation hook "certificates-relation-changed" (via hook dispatching script: dispatch) failed: exit status 1
```

charmcraft.yaml
```diff
@@ -26,6 +26,12 @@ bases:
...
 
+charm-libs:
+  - lib: "tls_certificates_interface.tls_certificates"
+    version: 3
+  - lib: "observability-libs.cert_handler"
+    version: 1
+
 parts:
   charm:
     plugin: charm
```
IbraAoad, line 37: I get this error when packing on charmcraft 3.2.1

```bash
charmcraft internal error: AttributeError("'int' object has no attribute 'partition'")
```
removing charm-libs and changing parts to be like below seems to fix the issue

```suggestion
parts:
  charm:
    charm-binary-python-packages:
      - "pydantic>=2"
      - "cryptography"
      - "jsonschema"
```

src/charm.py
ca-scribner: yeah you're right that we're going to get stuck in this situation regardless, and not be notified when we have action to resolve.  I opened #17 to capture this, and for now will just leave the current approach.

charmcraft.yaml
ca-scribner: Seems like charmcraft change how it lints the charm-libs version field between 3.1.2 and 3.2.1.  I believe it is now fixed, but let me know if it still doesn't work for you.  

background: [this charmcraft issue](https://github.com/canonical/charmcraft/issues/1954)

src/charm.py
ca-scribner: Yup good catch.  I refactored the pydantic models a bit, and it was exporting `None` for this field when it was unset.  To fix this, I've changed the model export to this:

```python
        return http_resource(
            metadata=ObjectMeta.from_dict(http_route.metadata.model_dump()),
            # Export without unset and None because None means nil in Kubernetes, which is not what we want.
            spec=http_route.spec.model_dump(exclude_unset=True, exclude_none=True),
        )
```

IbraAoad (approved):
Awesome job!!, thanks for covering non-tls communication in scenario tests as well.

PR #31, opened 2024-12-17
Create an Authorization Policy for any ingress relation

### LINKED ISSUE #30: Create an Authorization Policy for any ingress relation
### Enhancement Proposal

When an application relates to this charm to obtain a route through the ingress, traffic will be sent from the ingress workload directly to the related application.  If the related application has any existing Authorization Policies (either specific ones for that application, or a global allow-nothing), this ingress->application traffic will be blocked because no AuthorizationPolicies allow that traffic

To fix this, we need to create an AuthorizationPolicy for any application that relates to this ingress.

IbraAoad:
## Issue
Fixes #30 

## Solution
Create an L4 policy for each `httproute`.

## Testing Instructions

1. juju deploy istio-k8s
2. juju deploy istio-beacon with `model-on-mesh=true`
3. juju deploy istio-ingress
4. juju deploy alertmanager-k8s
5. juju relate alertmanager-k8s istio-ingress
6. An authorization policy should be created automatically and alertmanager-k8s ingress url should be reachable

## Out of PR scope
1- Generating policies for non-charmed workloads https://github.com/canonical/istio-ingress-k8s-operator/issues/32
2- Policies with multiple ports/paths for multi routes ingresses https://github.com/canonical/istio-ingress-k8s-operator/issues/23#issuecomment-2549390968

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -24,16 +24,19 @@ from lightkube.resources.apps_v1 import Deployment
 from lightkube.resources.core_v1 import Secret, Service
 from lightkube.types import PatchType
 from lightkube_extensions.batch import KubernetesResourceManager, create_charm_default_labels
-from ops import BlockedStatus, EventBase
+from ops import BlockedStatus, EventBase, main
 from ops.charm import CharmBase
-from ops.main import main
 from ops.model import ActiveStatus, MaintenanceStatus
 from ops.pebble import ChangeError, Layer
 
 # from lightkube.models.meta_v1 import Patch
 from models import (
     AllowedRoutes,
+    AuthorizationPolicyResource,
+    AuthorizationPolicySpec,
+    AuthRule,
     BackendRef,
+    From,
     GatewayTLSConfig,
     HTTPRequestRedirectFilter,
     HTTPRouteFilter,
@@ -46,10 +49,14 @@ from models import (
     Listener,
     Match,
     Metadata,
+    Operation,
     ParentRef,
     PathMatch,
     Rule,
     SecretObjectReference,
+    Source,
+    To,
+    WorkloadSelector,
 )
 
 logger = logging.getLogger(__name__)
@@ -68,6 +75,12 @@ RESOURCE_TYPES = {
     "ReferenceGrant": create_namespaced_resource(
         "gateway.networking.k8s.io", "v1beta1", "ReferenceGrant", "referencegrants"
     ),
+    "AuthorizationPolicy": create_namespaced_resource(
+        "security.istio.io",
+        "v1",
+        "AuthorizationPolicy",
+        "authorizationpolicies",
+    ),
 }
 
 GATEWAY_RESOURCE_TYPES = {RESOURCE_TYPES["Gateway"], Secret}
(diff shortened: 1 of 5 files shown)


src/models.py
```diff
@@ -166,3 +166,69 @@ class HTTPRouteResource(BaseModel):
 
     metadata: Metadata
     spec: HTTPRouteResourceSpec
+
+
+# Authrization Policy schema
+# Below is stripped down to cater for only L4 needed policies for ingress
+
+
```
dstathis, line 174: Not a blocking comment be any means, but we should probably make a lib of these at some point.

dstathis (approved):

src/charm.py
```diff
@@ -76,8 +89,10 @@
     RESOURCE_TYPES["ReferenceGrant"],
     RESOURCE_TYPES["HTTPRoute"],
 }
+AUTHORIZATION_POLICY_RESOURCE_TYPES = {RESOURCE_TYPES["AuthorizationPolicy"]}
 GATEWAY_LABEL = "istio-gateway"
 INGRESS_LABEL = "istio-ingress"
+AUTHORIZATION_POLICY_LABEL = "istio-ingress-authorization-policy"
```
ca-scribner, line 95: Can we rename this (and the two `*_LABEL`s above) to `AUTHORIZATION_POLICY_SPOPE`?  I just spent a few minutes writing about how these labels need to include `self.app.name`, etc, before realizing they're really just the scope and the labels themselves have more charm data :D

src/models.py
ca-scribner: Yeah.  I thought the same about `_get_peer_identity_for_juju_application()` which is now used in two charms.  a good thing to add to the backlog?

src/charm.py
```diff
@@ -439,6 +467,42 @@ def _construct_httproute(self, data: IngressRequirerData, prefix: str, section_n
             spec=http_route.spec.model_dump(exclude_none=True),
         )
 
+    def _construct_ingress_auth_policy(self, data: IngressRequirerData):
+
+        auth_policy = AuthorizationPolicyResource(
+            metadata=Metadata(
+                name=data.app.name + "-" + self.app.name + "-l4-policy",
```
ca-scribner, line 474: This can collide if the same app relates to two ingresses that have the same name but in different models.  If we deploy:
```
juju deploy -m model1 some-dashboard
juju deploy -m model2 istio-ingress
juju deploy -m model3 istio-ingress
```
then do the cross-model relations to connect everything, both istio-ingress charms will try to create the same `some-dashboard-istio-ingress-l4-policy` AP in the `model1` namespace. 

Adding the ingress's namespace in the AP name should prevent collision

src/charm.py
ca-scribner, line 1: have we tested this cross-model?  I think it should work but I'm not 100% sure
(this came up when thinking about [this comment](https://github.com/canonical/istio-ingress-k8s-operator/pull/31/files#r1889261514))

src/charm.py
ca-scribner: beacon [also needs to name policies](https://github.com/canonical/istio-beacon-k8s-operator/blob/abffe45a23a6ca72a30cc68fa8d8ef5a11ff4eda/src/charm.py#L446-L455).  Beacon needs to be more complex, but i wonder if we can format them similar-ish so they at least look like they came from the same group of applications.  

Looking at the inputs, I'm not sure its worth it, but mentioning it here in case you see a nice pattern that suits them both

ca-scribner (approved):
a few small suggestions, otherwise lgtm

src/charm.py
IbraAoad: Yep it should, the `istio-ingress` charm is designed to carry its identity even when not on the mesh, as long as we're deploying the policy in the ingressed app namespace so the policy can match the app with the selector then we're good.

src/charm.py
IbraAoad: I agree with appending the namespace, but I feel like implementing a similar naming to beacon would add unnecessary complexity since inputs are kinda always expected (i.e source app is always the ingress charm), also I don't think we will ever have collisions here since there's always 1-1 mapping between the ingress charm and target app

src/models.py
IbraAoad: https://github.com/canonical/istio-beacon-k8s-operator/issues/44

PR #47, opened 2025-03-13
Add an integration for registering authorization extension providers

### LINKED ISSUE #39: integrate with oauth2-proxy-k8s
### Enhancement Proposal

To enable authentication at the ingress, we need to integrate with  the [oauth2-proxy-k8s](https://github.com/canonical/oauth2-proxy-k8s-operator) charm to receive the required oauth2 details for integration.  

This should purely be us consuming a generic relation defined/owned by the oauth2-proxy-k8s charm.  No custom relations should be implemented.  

### Definition of Done

* istio-ingress can be integrated with the oauth2-proxy-k8s charm

### Out of scope

* actually implementing authentication on ingress.  That is covered by #40 

### Other background

See [Istio-Identity Platform Integration spec](https://docs.google.com/document/d/1D8sv_xOU5-tpXn3WGaLSl9WCsh29UfyrtuOu2up4oVU/edit?tab=t.0#heading=h.vux0i6b5dm8y) for more details

IbraAoad:
In tandem with https://github.com/canonical/istio-k8s-operator/pull/53

## Issue
Fixes #39 & #40 

## Context
In order for this ingress charm to be able to provide auth-d ingresses, it needs to be aware of the oauth provider configs (Service name & Port) as well as cascade those information to the `istio-k8s` charm

## Solution 
This PR adds the necessary integration between this charm and the `oauth2-proxy` charm via the `forward-auth` interface as well as the integration between this charm and `istio-k8s` via the `istio-ingress-config` interface

## Testing
1- juju deploy istio-k8s -m istio-system
2- juju deploy istio-ingress-k8s & oauth2-proxy-k8s -m ingress-auth
3- juju offer istio-k8s:istio-ingress-config ingress-config
4- juju consume admin.ingress-auth/ingress-config
5- juju relate istio-ingress-k8s ingress-config
6- juju relate istio-ingress-k8s:forward-auth oauth2-proxy-k8s
7- Validate the istiod has a new external auth provider 
8- Validate that a new auth policy is created in the ingress-auth model with empty rules (all traffic to go through relations)
9- Break any of the 2 relations and validate that 7&8 are deleted


src/charm.py
```diff
@@ -569,6 +642,29 @@ def _sync_all_resources(self):
             )
             self.on.refresh_certs.emit()
 
+    def _sync_ext_authz_config(self):
+        policy_manager = self._get_extz_auth_policy_resource_manager()
+        resources = []
+
+        auth_info = self.forward_auth.get_provider_info()
+        if not auth_info or not auth_info.decisions_address:
+            self.ingress_config.publish(ext_authz_service_name="fake_host", ext_authz_port="fake_port")
```
ca-scribner, line 651: See [here](https://github.com/canonical/istio-k8s-operator/pull/53/files#r2017445828) about generating the fake

src/charm.py
```diff
@@ -556,6 +628,7 @@ def _sync_all_resources(self):
                 )
                 return
             try:
+                self._sync_ext_authz_config()
```
ca-scribner, line 631: with auth in the mix, error handling here is really important.  If setting up auth ever fails, the charm needs to shut the ingress down with really high confidence.  Otherwise, auth config errors can lead to accidentally unauth'd resources.

Might not be directly applicable, but check out [istio-pilot](https://github.com/canonical/istio-operators/blob/b91873aa19a72a5492d8db2f7cb73ce2b1b806a3/charms/istio-pilot/src/charm.py#L313) for inspiration.  Note how it only sets `ingress_auth_reconcile_successful = True` when things worked, and it removes the gateway immediately after if auth was not successful.  We'll need something like that here

You're closer to this charm so give a think on how best to implement it here and we can discuss.  But something to that effect seems missing and important

src/charm.py
```diff
@@ -569,6 +642,29 @@ def _sync_all_resources(self):
...
+            return
+
+        parsed_url = urlparse(auth_info.decisions_address)
+        service_name = parsed_url.hostname
+        port = parsed_url.port
+
+        self.ingress_config.publish(ext_authz_service_name=service_name, ext_authz_port=str(port))
+        rel = self.model.get_relation(INGRESS_CONFIG_RELATION)
+        if rel and self.ingress_config.is_requirer_ready(rel):
+            provider_name = self.ingress_config.get_ext_authz_provider_name(rel)
```
ca-scribner, line 663: Say we have an ingress and we add forward_auth.  What is the UX on first reconcile here?  We'll enter this function, send ext_authz_* to istio-k8s, then silently fail this `if` statement.  Do we alert the user that things are only partly configured here?  I don't think we do, but we should

ca-scribner: This relates to [this comment](https://github.com/canonical/istio-ingress-k8s-operator/pull/47/files#r2017597895) too.  If someone tries to have auth (eg: `if len(auth_relations) > 0` then we:
* should only create a gateway if the auth is ready
* make it really clear to the user when auth is not ready, and why

In general, keep "fail closed" and a fully informed user in mind

ca-scribner: My gut says breaking `_sync_ext_auth_config` into a few helpers (one to manage auth_info, another to manage the istio-config relation, and maybe another to actually make the authz policy?) will make this easier to test and understand.  Just a suggestion though.  However you implement, keep in mind clarity is really important here since the cost of a mistake is accidentally exposing data

src/models.py
```diff
@@ -222,9 +238,20 @@ class AuthRule(BaseModel):
...
+    selector: Optional[WorkloadSelector] = Field(default=None)
+    provider: Optional[Provider] = Field(default=None)
+
+    @model_validator(mode="after")
+    def validate_target(self):
+        """Validate that at most one of targetRefs and selector is defined."""
+        if self.targetRefs is not None and self.selector is not None:
+            raise ValueError("At most one of targetRefs and selector can be set")
+        if self.provider is not None and self.action is not Action.custom:
+            raise ValueError("CUSTOM action must be set when specifying extension providers")
```
ca-scribner, line 253: Based on the docstring, this seems out of scope for this validator.  Should this be a separate validator?

ca-scribner: also, I can follow the logic here, but I don't have a good sense of how validators work.  Some unit test coverage proving this out would really help

ca-scribner (commented):
Generally I'm +1 on the direction.  Only real concerns are around ensuring we never fail open when auth is misconfigured (eg: if the user ever expresses the desire for auth, we never allow unauth'd access)

lib/charms/istio_k8s/v0/istio_ingress_config.py
```diff
@@ -151,13 +166,23 @@ def publish(
             databag.update(data)
             log.debug("Published provider data: %s to relation: %s", data, relation)
 
-    def get_ext_authz_provider_name(self, relation: Relation) -> Optional[str]:
+    def publish_fake_authz_config(self) -> None:
+        """Publish a fake external authorization configuration to all related applications."""
```
ca-scribner, line 170: Once you have the bug open, lets add a TODO or something here that explains why we have a fake publisher and the conditions for when we can remove it (eg: when juju handles this properly)

lib/charms/istio_k8s/v0/istio_ingress_config.py
```diff
@@ -245,6 +270,21 @@ def get_provider_ext_authz_info(
             log.debug("Failed to validate provider data: %s", e)
             return None
 
+    def is_fake_authz_config(self, relation: Relation) -> bool:
+        """Check if the provider has published a fake external authorization configuration.
```
ca-scribner, line 274: Same suggestion as [here](https://github.com/canonical/istio-ingress-k8s-operator/pull/47/files/2d7ac060df6751700e15ca6d23fe139d8bb77e2a..f5cb101387a89342171593cbe1d8375e8bde732c#r2029110406)

src/charm.py
```diff
@@ -617,53 +623,148 @@ def _construct_redirect_to_https_httproute(
...
+        Flow:
+        1. Check forward authentication configuration.
+            - If auth relation exists but no decisions address, set to blocked and remove gateway.
+        2. Synchronize gateway resources and validate readiness.
+        3. Validate the external hostname.
+        4. Synchronize external authorization configuration.
+            - If missing valid ingress-config relation when auth is provided, set to blocked and remove gateway.
+        5. Synchronize ingress resources and set up the proxy service.
+        6. Request certificate inspection.
+        """
```
ca-scribner, line 637: :heart:

src/charm.py
```diff
@@ -617,53 +623,148 @@ def _construct_redirect_to_https_httproute(
...
+                includes a decisions_address.
+            - decisions_address is the retrieved decisions address if available;
+                otherwise, None.
+
+        If a forward-auth relation exists but is missing the decisions address,
+        the configuration is considered not ready.
+        """
+        if len(self.model.relations) == 0:
+            logger.debug("No relations found. Configuration is considered ready.")
+            return True, None
```
ca-scribner, line 704: I'd take this out.  Its correct this is a valid escape, but I don't think it saves any time - we can just try to get the data like we do below

src/charm.py
```diff
@@ -545,29 +623,148 @@ def _construct_redirect_to_https_httproute(
...
+        3. Validate the external hostname.
+        4. Synchronize external authorization configuration.
+            - If missing valid ingress-config relation when auth is provided, set to blocked and remove gateway.
+        5. Synchronize ingress resources and set up the proxy service.
+        6. Request certificate inspection.
+        """
+        # 1. Validate authentication configuration.
+        auth_ready, auth_decisions_address = self._check_forward_auth_configuration()
+        # Guard if We have an auth relation but provider didn't publish an auth_decisions_address
+        if not auth_ready:
```
ca-scribner, line 641: This reads "if auth is not ready", but what it means is "if auth **is configured but** not ready".  

The trouble here is `auth_ready` is True both cases of "configured and ready" and "never configured".  See comments around `_check_forward_auth_configuration` for some suggestions about this

src/charm.py
```diff
@@ -617,53 +623,148 @@ def _construct_redirect_to_https_httproute(
...
+        """
+        if len(self.model.relations) == 0:
+            logger.debug("No relations found. Configuration is considered ready.")
+            return True, None
+
+        relation = self.model.get_relation(FORWARD_AUTH_RELATION)
+        if not relation or not relation.app:
+            logger.debug(
+                "Forward-auth relation or it's associated app is missing. Configuration is considered ready."
+            )
```
ca-scribner, line 710: Not a hard blocker since this is a debug comment, but as a user if I saw this I'd be confused.  `X is missing, configuration is considered ready` isn't easy to understand without understanding of the charm inners.   

I think this helper can be refactored to where you don't need this comment anyway, but if that isn't true then we should think about rewording

src/charm.py
```diff
@@ -617,53 +623,148 @@ def _construct_redirect_to_https_httproute(
...
+            - If auth relation exists but no decisions address, set to blocked and remove gateway.
+        2. Synchronize gateway resources and validate readiness.
+        3. Validate the external hostname.
+        4. Synchronize external authorization configuration.
+            - If missing valid ingress-config relation when auth is provided, set to blocked and remove gateway.
+        5. Synchronize ingress resources and set up the proxy service.
+        6. Request certificate inspection.
+        """
+        # 1. Validate authentication configuration.
+        auth_ready, auth_decisions_address = self._check_forward_auth_configuration()
```
ca-scribner, line 639: `forward_auth` as a name doesn't really have meaning in istio-ingress - let's not name our helpers that specifically.  Instead, maybe its `_check_auth_configuration`?

src/charm.py
ca-scribner: Also, I was 30 min into this PR before I realized `auth_ready` is just that the data from **forward-auth** is ready, not that we also have the istio-k8s external authorizer configured and ready.  I read `auth_ready` as *everything* is ready for auth, not just a piece

src/charm.py
```diff
@@ -617,53 +623,148 @@ def _construct_redirect_to_https_httproute(
...
-            return
+        ingress_ready = self.ingress_config.is_requirer_ready()
+        auth_valid = decisions_address is not None
+
+        if ingress_ready:
+            if auth_valid:
+                # Scenario 1: Valid ingress and valid auth relation.
+                self.publish_ext_authz_config(decisions_address)
+                provider_name = self.ingress_config.get_ext_authz_provider_name()
+                resources.append(self._construct_ext_authz_policy(provider_name))  # type: ignore
```
ca-scribner, line 749: Think there's a bug here.  

On first pass through, we will `publish_ext_authz_config` which sends ext auth config to istio.  We'll then immediately do `self.ingress_config.get_ext_authz_provider_name()`, which will be unpopulated (because istio hasn't responded yet) so I think it'll mean `provider_name=None`, then we do `self._construct_ext_authz_policy(None)` which I think fails

I wonder if there's a better approach here.  The main reason I think that is that is because calling `self.publish_ext_authz_config(decisions_address)` with a new `decisions_address` invalidates the data in `self.ingress_config`.  With that said though, maybe this just has to be coded in a way that expects `self.ingress_config.get...` might return None, which results in things not being ready, and then a later reconcile fixes it?

ca-scribner: different bug... If `self.publish_ext_authz_config()` is what sends the auth config onto the `self.ingress_config` relation, but `self.publish_ext_authz_config()` is behind an `ingress_ready` guard, how can ingress_config ever be configured?

src/charm.py
```diff
@@ -617,53 +623,148 @@ def _construct_redirect_to_https_httproute(
...
+            return False, None
+
+        return True, auth_info.decisions_address
+
+    def _sync_ext_authz_config(self, decisions_address: Optional[str]) -> bool:
+        """Synchronize external authorization configuration.
+
+        Scenarios:
+        1. Valid ingress and valid auth: publish actual service, construct policy, return True.
+        2. Valid ingress and no auth: publish fake service, reconcile empty resources, return True.
```
ca-scribner, line 731: Anywhere we discuss "fake" data, I think we should instead say clear the data but then have a separate note saying "there is a bug, instead of clearing the data we publish fake data bla bla bla".  That way anyone reading this understands what is meant by fake data.

src/charm.py
```diff
@@ -617,53 +623,148 @@ def _construct_redirect_to_https_httproute(
...
+        3. No ingress and valid auth: do nothing, reconcile empty resources, return False.
+        4. No ingress and no auth: do nothing, reconcile empty resources, return True.
+
+        Returns:
+            True for Scenarios 1, 2, 4 and False for Scenario 3.
```
ca-scribner, line 736: I find the return of this function confusing.  It might be less confusing if described differently.  Saying "returns True for 1, 2, 4 and False for 3" doesn't really tell me what True *means*.  Does True mean "successfully synced"?  I dont think it does, since in all cases a reconcile happens (just in some cases it reconciles to what you want, others it reconciles to empty for protection).

I'm not sure how to refactor this for better clarity.  Is this function doing too much?  We should discuss

ca-scribner (commented):
Generally like the direction but I had a hard time following some bits.  I think we can simplify this through a little refactoring to make the business logic clearer

src/charm.py
```diff
@@ -545,29 +623,128 @@ def _construct_redirect_to_https_httproute(
...
+    def _publish_ext_authz_config(self, decisions_address: Optional[str]):
+        """Publish the external authorization service configuration using the provided decisions_address."""
+        if not decisions_address:
+            self.ingress_config.clear()
+            return
+
+        parsed_url = urlparse(decisions_address)
+        service_name = parsed_url.hostname
+        port = parsed_url.port
+        self.ingress_config.publish(ext_authz_service_name=service_name, ext_authz_port=str(port))
```
ca-scribner, line 737: Lets add a note here about how this needs a leader guard, but that we might need to do more to make this work for scaling anyway.  I say add a note rather than a guard here because I'm not sure if we want the guard here or somewhere else, but if you're confident you can just add the guard itself instead

ca-scribner (commented):
this is really close.  It reads much easier than before which is great!  Had a few comments but pretty small stuff

src/charm.py
```diff
@@ -545,29 +623,128 @@ def _construct_redirect_to_https_httproute(
...
-                self.unit.status = BlockedStatus("Issue with setting up an ingress")
-
-            # Request a cert refresh in case configuration has changed
-            # The cert handler will only refresh if it detects a meaningful change
-            logger.info(
-                "Requesting CertHandler inspect certs to decide if our CSR has changed and we should re-request"
+        # 3. Validate external hostname.
+        if not self._external_host:
+            self.unit.status = BlockedStatus(
+                "Invalid hostname provided, Please ensure this adheres to RFC 1123."
```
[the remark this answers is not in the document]
IbraAoad: It's behind a `_is_valid_hostname` guard when used before this check, if invalid the gateway will be created without a host and afterwards we set the charm to blocked here.

I think it might be worth it to delete the gateway at this stage like we're doing when auth is misconfigured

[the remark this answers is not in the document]
ca-scribner: k I'm good with whatever you decide on then!

src/charm.py
```diff
@@ -282,6 +324,10 @@ def _on_ingress_data_removed(self, _):
         """Handle a unit removing the data needed to provide ingress."""
         self._sync_all_resources()
 
+    def _remove_gateway_resources(self):
+        kgm = self._get_gateway_resource_manager()
+        kgm.delete()
```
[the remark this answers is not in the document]
ca-scribner: I'm not gonna block on this, but I'm reopening because I can't tell if you missed it or just really like keeping these separate :D

src/charm.py
```diff
@@ -545,29 +623,111 @@ def _construct_redirect_to_https_httproute(
...
+        if self.model.get_relation(FORWARD_AUTH_RELATION) and not auth_decisions_address:
+            self.unit.status = BlockedStatus(
+                "Authentication configuration incomplete; ingress is disabled."
+            )
+            self._remove_gateway_resources()
+            return
+
+        # 2. Publish or clear the auth_decisions_address in ingress-config, if related.
+        if self.model.get_relation(INGRESS_CONFIG_RELATION):
+            self._publish_ext_authz_config(auth_decisions_address)
```
ca-scribner, line 650: We should move this up.  Consider if:
* we have oauth relation and received decision address
* we have istio-config relation and have sent the decision address

Then the oauth charm updates the oauth relation to change decision_address to None.  Our charm here will Block(Authentication configuration incomplete) **and return** rather than update istio_k8s.  So a disabled oauth provider will still be around

We can fix this by ensuring this charm publishes data to istio-k8s regardless of whatever else happens

src/charm.py
```diff
@@ -545,29 +623,111 @@ def _construct_redirect_to_https_httproute(
...
+            return
+
+        parsed_url = urlparse(decisions_address)
+        service_name = parsed_url.hostname
+        port = parsed_url.port
+        # TODO: Below probably needs to be leader guarded
+        # we should think about this as part of working on #issues/16
+        self.ingress_config.publish(ext_authz_service_name=service_name, ext_authz_port=str(port))
+
+    def _sync_ext_authz_config(self, auth_decisions_address: Optional[str]):
```
ca-scribner, line 722: Just a suggestion: when I read this code I often dont realize the idea is sending auth config to istio-k8s.  "sync" could be sending data out, or updating our objects based on the data we received.  Maybe the name could be more explicit?  
* `_publish_ext_authz_config`
* `_publish_to_istio_ingress_config_relation`?

Ignore if you disagree

tests/scenario/test_auth.py
```diff
@@ -0,0 +1,126 @@
...
+    mock_is_ready,
+    mock_setup,
+    mock_publish,
+    mock_sync,
+    forward_auth_present,
+    expected_decision,
+    istio_ingress_charm,
+    istio_ingress_context,
+):
+    relations = []
```
ca-scribner, line 42: I love that you have this explanation.  My only suggestion is to bring it down into the decorators.  Reason being that I can easily see someone reading the cases or patches and just not noticing this comment up here.

Instead, maybe:

```suggestion
@pytest.mark.parametrize(
    "forward_auth_present, expected_decision",
    [
        # When a forward-auth relation exists, its decisions address should be used
        (True, "http://auth-service:80"),
        # When no forward-auth relation exists, the decision address should be None, which would then clear the ingress-config databag
        (False, None),  # No forward-auth relation.
    ],
)
# to verify it is called
@patch.object(IstioIngressCharm, "_sync_ext_authz_config")
# to verify it is called
@patch.object(IstioIngressCharm, "_publish_ext_authz_config")
# to skip side effects
@patch.object(IstioIngressCharm, "_setup_proxy_pebble_service")
# so that gateway readiness passes
@patch.object(IstioIngressCharm, "_is_ready", return_value=True)
def test_ext_authz_setup(
    mock_is_ready,
    mock_setup,
    mock_publish,
    mock_sync,
    forward_auth_present,
    expected_decision,
    istio_ingress_charm,
    istio_ingress_context,
):
   """Test external authorization configuration setup for when we have a valid ingress-config and different forward_auth situations.""":

    relations = []
```

tests/scenario/test_auth.py
```diff
@@ -0,0 +1,126 @@
...
+    mock_publish.assert_called_once_with(expected_decision)
+    mock_sync.assert_called_once_with(expected_decision)
+    assert isinstance(out.unit_status, ActiveStatus)
+    assert out.unit_status.message.startswith("Serving at")
+
+
+# Test external authorization configuration setup when the forward-auth or ingress-config relation is incomplete or missing:
+#   - Scenario A: The forward-auth relation is present but provides no decisions address.
+#   - Scenario B: The forward-auth relation provides a decisions address, but the ingress-config relation exists with incomplete data.
+# In both scenarios, the charm is expected to set a BlockedStatus and call _remove_gateway_resources.
```
ca-scribner, line 74: same comment as above.  LOVE the detail, but I'd shift it down lower.  You've already got most of the detail down there already

ca-scribner (approved):
lgtm!  great work

added a few minor nits/suggestions you can consider, but all looks great

PR #55, opened 2025-04-15
Pass a list of ingressed endpoints to the oauth2-proxy charm when connected over forward-auth

IbraAoad:
## Issue
The [oauth2-proxy](https://github.com/canonical/oauth2-proxy-k8s-operator) charm expects a list of ingress-ed endpoints when connected over the `forward-auth` interface, this list will be used later on by `oauth` to validate which apps should can request authorization.

## Solution
This PR adds a new step in the reconcile flow to extract the list of ingressed apps from `proxied_endpoints` and sends it over the forward-auth relation.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -12,7 +12,7 @@ from typing import Any, Dict, Optional, cast
 from urllib.parse import urlparse
 
 from charms.istio_k8s.v0.istio_ingress_config import IngressConfigProvider
-from charms.oauth2_proxy_k8s.v0.forward_auth import ForwardAuthRequirer
+from charms.oauth2_proxy_k8s.v0.forward_auth import ForwardAuthRequirer, ForwardAuthRequirerConfig
 from charms.observability_libs.v1.cert_handler import CertHandler
 from charms.prometheus_k8s.v0.prometheus_scrape import MetricsEndpointProvider
 from charms.tempo_coordinator_k8s.v0.charm_tracing import trace_charm
@@ -634,7 +634,8 @@ class IstioIngressCharm(CharmBase):
         5. Synchronize gateway resources and validate readiness.
         6. Validate the external hostname.
         7. Synchronize ingress resources and set up the proxy service.
-        8. Request certificate inspection.
+        8. Update forward auth relation data with ingressed apps.
+        9. Request certificate inspection.
         """
         # 1. Check authentication configuration.
         auth_decisions_address = self._get_oauth_decisions_address()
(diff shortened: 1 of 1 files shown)


src/charm.py
```diff
@@ -683,7 +684,14 @@ def _sync_all_resources(self):
             self.unit.status = BlockedStatus("Issue with setting up an ingress")
             return
 
-        # 8. Request certificate inspection.
+        # 8. Update forward auth relation data with ingressed apps.
+        if self.model.get_relation(FORWARD_AUTH_RELATION):
+            ingressed_apps = list(self.ingress_per_appv2.proxied_endpoints.keys())
```
IbraAoad, line 689:
```suggestion
    def _get_ingressed_apps(self) -> List[str]:
        """Return a list of application names for which ingress is configured and active."""
        results= []

        for ingress_relation in self.ingress_per_appv2.relations:
            if not ingress_relation.app:
                logger.warning(
                    f"no app in relation {ingress_relation} when fetching proxied endpoints: skipping"
                )
                continue
            try:
                ingress_data = self.ingress_per_appv2._published_url(ingress_relation)
            except NotReadyError:
                logger.warning(
                    f"no published url found in {ingress_relation}: "
                    f"istio ingress didn't publish_url yet to this relation: skipping"
                )
                continue

            if not ingress_data:
                logger.warning(f"relation {ingress_relation} not ready yet.")
                continue
            results.append(ingress_relation.app.name)
        return results
```

We can do this if anyone is opposed to using `proxied_endpoints` directly, it's basically a stripped down version of `proxied_endpoints`, but I don't think it's worth the duplication.

src/charm.py
```diff
@@ -683,7 +684,14 @@ def _sync_all_resources(self):
...
+            self.forward_auth.update_requirer_relation_data(
+                ForwardAuthRequirerConfig(ingress_app_names=ingressed_apps)
+            )
```
ca-scribner, line 692: We set this as None in our __init__, then we configure it down here.  Does that mean __init__ nulls out the data sent every time the charm starts?  I'm worrying about if on every charm execution we tell oauth2 "we deleted everything" and then a moment later send "now we've updated it to X again"

ca-scribner: Similarly, if the charm exits early, do we (or should we?) null out this data, or just leave it however it was previously?

src/charm.py
ca-scribner: I'm not 100% sure what we are supposed to send here, but I think the work i'm doing with the split ingress will produce a list of apps that'll work here so i'm not too concerned about using the proxied endpoints for now

src/charm.py
ca-scribner: I'm resolving this - the only time we send data is when calling this `update_requirer_relation_data` function so what I was worried about doesn't matter

ca-scribner (approved):

PR #58, opened 2025-04-17
refactor: refactor how routes are generated for the ingress relation

ca-scribner:
## Issue
In future, we plan to offer two ingress relations (ingress and ingress_unauthenticated).  To facilitate that future change, this commit refactors how routes are generated for the current single ingress relation.  This commit should have no outward functional change.

This refactor tries to address the following points:
* while previously it was impossible for two applications to request the same route (because each application could only relate to us once, and we generated routes by ('/model-name-app-name'), duplicates are possible once we add a second ingress relation (the same app could relate to both ingress and ingress-unauthenticated).  In future if we support user-specificed paths in ingress v3, we'll face this problem as well.  Thus, duplicate detection has been added here
* previously, we supported a single ingress version (v2) on one relation endpoint.  As we expand this to a second endpoint and later ingress v3, we'll need to support different handlers for different related applications

## Context
.

## Testing Instructions
CI should give full coverage afaict

## Upgrade Notes
.


ca-scribner:
This PR implements a somewhat functional approach, but I wonder if it would be more maintainable/easier to read we created some sort of RouteManager class that encapsulated the getting route data from ingress / sending things to those ingressed apps.  It would be a lot of the same code, but no longer coupled to the charm object

src/charm.py
```diff
@@ -950,5 +1085,74 @@ def _get_peer_identity_for_juju_application(app_name, namespace):
...
 
+def get_relation_by_name_and_app(relations, remote_app_name):
+    """Return the relation object associated with a given remote app."""
+    for rel in relations:
+        if rel.app.name == remote_app_name:
+            return rel
+    raise KeyError(f"Could not find relation with remote_app_name={remote_app_name}")
+
+
+def deduplicate_app_route_data(application_route_data) -> bool:
```
IbraAoad, line 1096: Is this function necessary for this pass?, as of now we don't really have a way for a charm to request a prefix and it always defaults to `{model_name}-{app_name}' so there's never going to be a collision.

However i'm with you on the fact that once we have `ingressV3` this will become an issue, but I think we should talk about how we want to approach it, this implementation if I understand it correctly removes all colliding ingresses requesting the same path and I think a better way to do it, is to honor which charm requested the prefix first and refuse the other ingress.

I also think all the work should be part of #57 and deferred until we have an interface that offers the feature.

tests/scenario/test_ingress.py
```diff
@@ -40,32 +41,38 @@ def mock_ingress_requirer_data():
     ],
 )
 def test_construct_httproute_with_strip_prefix(
-    strip_prefix, filters, istio_ingress_charm, istio_ingress_context, mock_ingress_requirer_data
+    strip_prefix, filters, istio_ingress_charm, istio_ingress_context
```
IbraAoad, line 44: I think we can drop the `mock_ingress_requirer_data` fixture implementation since we're passing everything explicitly now.

IbraAoad:
> This PR implements a somewhat functional approach, but I wonder if it would be more maintainable/easier to read we created some sort of RouteManager class that encapsulated the getting route data from ingress / sending things to those ingressed apps. It would be a lot of the same code, but no longer coupled to the charm object


+1 on the suggestion, will make the charm code a lot easier to follow and I think probably easier to self-test as well.

src/charm.py
IbraAoad: > duplicates are possible once we add a second ingress relation (the same app could relate to both ingress and ingress-unauthenticated). 

I'm not sure this is possible now without `v3`, I think we should actually block/or neglect the second request with an error if the same app related to us on both interfaces, if it's 2 deployments of the same app on 2 different models then we don't have an issue.

If I'm saying words that don't make any sense, it's probably cause I'm confused and my head is stuck on the idea that we'll never have collisions as long as `v3` isn't here, even with the ingress split.

ca-scribner: Yes you're right that in this PR (ingress v2, single ingress endpoint) duplicates are not possible.  Happy to remove this for now

... but :D if we remove this here, we'll have to include it instead in #60.  In #60, which adds the second ingress endpoint for unauth'd traffic, the same application can relate to us twice (once on each of the ingress relations) and that will be a duplicate that we need to handle

ca-scribner: So I'm happy to move it to the other PR if that makes more sense.  But removing it entirely I'm not sure about

ca-scribner: Better messaging can definitely be part of #57, but imo we still have to handle it gracefully here.  I don't see how we can prevent someone from relating on both endpoints.  

A (bit contrived) example of this is [in #60's ipa-tester](https://github.com/canonical/istio-ingress-k8s-operator/blob/a5997f3a32b535dd6f5e6f630f2fe97e1646bf53/tests/integration/testers/ipa/src/charm.py#L17).  Though I doubt someone could accidentally do this

src/charm.py
```diff
@@ -716,6 +762,84 @@ def _get_oauth_decisions_address(self) -> Optional[str]:
...
+            if key not in application_route_data:
+                application_route_data[key] = {"handler": relation_handler, "routes": []}
+
+            if not rel.active or not relation_handler.is_ready(rel):
+                # No active routes for this related application
+                continue
+
+            data = relation_handler.get_data(rel)
+            # strip_prefix = None equates to False
+            data.app.strip_prefix = data.app.strip_prefix or False
```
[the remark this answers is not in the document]
ca-scribner: Fixed this in the next commit.  Had to do a few other changes with it

PR #61, opened 2025-04-22
feat: enable scaling of the charm and underlying Gateway workload

### LINKED ISSUE #16: Istio Ingress not behave properly when scaled
### Bug Description

When Istio Ingress is scaled beyond 1, It goes in to error state. It should instead continue to work and it should scale custom deployments accordingly.

### To Reproduce

juju scale-application istio-ingress-k8s 2

### Environment

latest/edge

### Relevant log output

```shell
unit-istio-ingress-k8s-1: 23:13:37 ERROR unit.istio-ingress-k8s/1.juju-log Uncaught exception while in charm code:
Traceback (most recent call last):
  File "/var/lib/juju/agents/unit-istio-ingress-k8s-1/charm/./src/charm.py", line 374, in <module>
    main(IstioIngressCharm)
  File "/var/lib/juju/agents/unit-istio-ingress-k8s-1/charm/venv/ops/main.py", line 551, in main
    manager.run()
  File "/var/lib/juju/agents/unit-istio-ingress-k8s-1/charm/venv/ops/main.py", line 530, in run
    self._emit()
  File "/var/lib/juju/agents/unit-istio-ingress-k8s-1/charm/venv/ops/main.py", line 519, in _emit
    _emit_charm_event(self.charm, self.dispatcher.event_name)
  File "/var/lib/juju/agents/unit-istio-ingress-k8s-1/charm/venv/ops/main.py", line 147, in _emit_charm_event
    event_to_emit.emit(*args, **kwargs)
  File "/var/lib/juju/agents/unit-istio-ingress-k8s-1/charm/venv/ops/framework.py", line 348, in emit
    framework._emit(event)
  File "/var/lib/juju/agents/unit-istio-ingress-k8s-1/charm/venv/ops/framework.py", line 860, in _emit
    self._reemit(event_path)
  File "/var/lib/juju/agents/unit-istio-ingress-k8s-1/charm/venv/ops/framework.py", line 950, in _reemit
    custom_handler(event)
  File "/var/lib/juju/agents/unit-istio-ingress-k8s-1/charm/./src/charm.py", line 135, in _on_config_changed
    self._sync_all_resources()
  File "/var/lib/juju/agents/unit-istio-ingress-k8s-1/charm/./src/charm.py", line 291, in _sync_all_resources
    self._sync_ingress_resources()
  File "/var/lib/juju/agents/unit-istio-ingress-k8s-1/charm/./src/charm.py", line 308, in _sync_ingress_resources
    raise RuntimeError("Ingress can only be provided on the leader unit.")
RuntimeError: Ingress can only be provided on the leader unit.
```

### Additional context

_No response_

IbraAoad:
## Issue
Fixes #16 

## Solution
This PR adds a HorizontalPodAutoscaler that is bound to the charm and would always update the gateway's deployment replicas to match the charm's unit count

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -331,10 +331,22 @@ class IstioIngressCharm(CharmBase):
         self._sync_all_resources()
 
     def _on_remove(self, _):
-        """Event handler for remove."""
+        """Event handler for remove.
+
+        The objective of this handler is to remove all application-scoped resources when the application is being scaled
+        to 0 or removed.  We intentionally do not put this removal action behind a leader guard (eg, behind
+        `if self.unit.is_leader()`) for the reasons discussed
+        [here](https://github.com/canonical/istio-ingress-k8s-operator/issues/16).
+        """
         # if there are still units left, skip removal
         if self.model.app.planned_units() > 0:
+            logger.info(
+                "Handling remove event: skipping resource removal because application is not scaling to 0."
+            )
             return
+        logger.info(
+            "Handling remove event: Attempting to remove application resources because application is scaling to 0."
+        )
 
         # Removing tailing ingresses
         kim = self._get_ingress_route_resource_manager()
@@ -898,25 +910,24 @@ class IstioIngressCharm(CharmBase):
 
     def _sync_gateway_resources(self):
         unit_count = self.model.app.planned_units()
-
-        # if there are no units left (unit_count < 1), skip reconcile:
-        #  - we assume this is a remove event and the remove hook will clean everything up
-        #  - updating the HPA with zero replicas is invalid
-        if unit_count < 1:
-            return
-
         krm = self._get_gateway_resource_manager()
         resources_list = []
         tls_secret_name = None
 
-        if secret := self._construct_gateway_tls_secret():
-            resources_list.append(secret)
-            if secret.metadata is None:
-                raise ValueError("Unexpected error: secret.metadata is None")
-            tls_secret_name = secret.metadata.name
+        # Skip reconciliation if no units are left (unit_count < 1):
+        #  - This typically indicates an application removal event; we rely on the remove hook for cleanup.
+        #  - Attempting to reconcile with an HPA that sets replicas to zero is invalid.
+        #  - This guard exists because some events can call _sync_all_resources before the remove hook runs,
+        #    leading to k8s validation webhook errors when planned_units is 0.
+        if unit_count > 0:
+            if secret := self._construct_gateway_tls_secret():
+                resources_list.append(secret)
+                if secret.metadata is None:
+                    raise ValueError("Unexpected error: secret.metadata is None")
+                tls_secret_name = secret.metadata.name
 
-        resources_list.append(self._construct_gateway(tls_secret_name=tls_secret_name))
-        resources_list.append(self._construct_hpa(unit_count))
+            resources_list.append(self._construct_gateway(tls_secret_name=tls_secret_name))
+            resources_list.append(self._construct_hpa(unit_count))
 
         krm.reconcile(resources_list)
 
--- a/src/charm.py
+++ b/src/charm.py
@@ -319,7 +319,8 @@ class IstioIngressCharm(CharmBase):
 
     def _on_remove(self, _):
         """Event handler for remove."""
-        if not self.unit.is_leader():
+        # if there are still units left, skip removal
+        if self.model.app.planned_units() > 0:
             return
 
         # Removing tailing ingresses
@@ -774,8 +775,15 @@ class IstioIngressCharm(CharmBase):
         policy_manager.reconcile(resources)
 
     def _sync_gateway_resources(self):
-        krm = self._get_gateway_resource_manager()
         unit_count = self.model.app.planned_units()
+
+        # if there are no units left (unit_count < 1), skip reconcile:
+        #  - we assume this is a remove event and the remove hook will clean everything up
+        #  - updating the HPA with zero replicas is invalid
+        if unit_count < 1:
+            return
+
+        krm = self._get_gateway_resource_manager()
         resources_list = []
         tls_secret_name = None
 
--- a/src/charm.py
+++ b/src/charm.py
@@ -105,7 +105,6 @@ GATEWAY_SCOPE = "istio-gateway"
 INGRESS_SCOPE = "istio-ingress"
 INGRESS_AUTH_POLICY_SCOPE = "istio-ingress-authorization-policy"
 EXTZ_AUTH_POLICY_SCOPE = "external-authorizer-authorization-policy"
-HPA_SCOPE = "istio-ingress-hpa"
 
 INGRESS_CONFIG_RELATION = "istio-ingress-config"
 FORWARD_AUTH_RELATION = "forward-auth"
@@ -698,7 +697,6 @@ class IstioIngressCharm(CharmBase):
         * Update forward auth relation data with ingressed apps.
         * Request certificate inspection.
         """
-
         if not self.unit.is_leader():
             self.unit.status = ActiveStatus("Backup unit; standing by for leader takeover")
             return
--- a/src/charm.py
+++ b/src/charm.py
@@ -94,14 +94,13 @@ RESOURCE_TYPES = {
     ),
 }
 
-GATEWAY_RESOURCE_TYPES = {RESOURCE_TYPES["Gateway"], Secret}
+GATEWAY_RESOURCE_TYPES = {RESOURCE_TYPES["Gateway"], Secret, HorizontalPodAutoscaler}
 INGRESS_RESOURCE_TYPES = {
     RESOURCE_TYPES["GRPCRoute"],
     RESOURCE_TYPES["ReferenceGrant"],
     RESOURCE_TYPES["HTTPRoute"],
 }
 AUTHORIZATION_POLICY_RESOURCE_TYPES = {RESOURCE_TYPES["AuthorizationPolicy"]}
-HPA_RESOURCE_TYPES = {HorizontalPodAutoscaler}
 
 GATEWAY_SCOPE = "istio-gateway"
 INGRESS_SCOPE = "istio-ingress"
@@ -191,8 +190,8 @@ class IstioIngressCharm(CharmBase):
             self.on[INGRESS_CONFIG_RELATION].relation_broken, self._handle_ingress_config
         )
         self.framework.observe(self.on.leader_elected, self._handle_ingress_config)
-        self.framework.observe(self.on[PEERS_RELATION].relation_joined, self._on_peers_changed)
         self.framework.observe(self.on[PEERS_RELATION].relation_changed, self._on_peers_changed)
+        self.framework.observe(self.on[PEERS_RELATION].relation_departed, self._on_peers_changed)
 
         # During the initialisation of the charm, we do not have a LoadBalancer and thus a LoadBalancer external IP.
         # If we need that IP to request the certs, disable cert handling until we have it.
@@ -294,14 +293,6 @@ class IstioIngressCharm(CharmBase):
             logger=logger,
         )
 
-    def _get_hpa_resource_manager(self):
-        return KubernetesResourceManager(
-            labels=create_charm_default_labels(self.app.name, self.model.name, scope=HPA_SCOPE),
-            resource_types=HPA_RESOURCE_TYPES,  # pyright: ignore
-            lightkube_client=self.lightkube_client,
-            logger=logger,
-        )
-
     def _on_cert_handler_cert_changed(self, _):
         """Event handler for when tls certificates have changed."""
         self._sync_all_resources()
@@ -334,9 +325,6 @@ class IstioIngressCharm(CharmBase):
         # Removing tailing ingresses
         kim = self._get_ingress_resource_manager()
         kim.delete()
-
-        khm = self._get_hpa_resource_manager()
-        khm.delete()
 
         self._remove_gateway_resources()
 
@@ -675,10 +663,9 @@ class IstioIngressCharm(CharmBase):
         * If auth relation exists but no decisions address, set to blocked and remove gateway.
         * Synchronize external authorization configuration.
             - If missing valid ingress-config relation when auth is provided, set to blocked and remove gateway.
-        * Synchronize gateway resources and validate readiness.
+        * Reconcile HPA and gateway resources to align replicas with unit count and ensure gateway readiness.
         * Validate the external hostname.
         * Synchronize ingress resources and set up the proxy service.
-        * Reconcile HPA so Gateway deployment's replicas match the current unit count
         * Update forward auth relation data with ingressed apps.
         * Request certificate inspection.
         """
@@ -709,7 +696,7 @@ class IstioIngressCharm(CharmBase):
             return
         self._sync_ext_authz_auth_policy(auth_decisions_address)
 
-        # Synchronize gateway resources and check readiness.
+        # Reconcile HPA and gateway resources
         self._sync_gateway_resources()
         self.unit.status = MaintenanceStatus("Validating gateway readiness")
         if not self._is_ready():
@@ -731,9 +718,6 @@ class IstioIngressCharm(CharmBase):
             logger.error("Ingress sync failed: %s", e)
             self.unit.status = BlockedStatus("Issue with setting up an ingress")
             return
-
-        # Reconcile HPA so Gateway deployment's replicas match the current unit count
-        self._sync_hpa_resources()
 
         # Update forward auth relation data with ingressed apps.
         if self.model.get_relation(FORWARD_AUTH_RELATION):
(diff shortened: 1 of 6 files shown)


src/charm.py
```diff
@@ -684,14 +732,17 @@ def _sync_all_resources(self):
             self.unit.status = BlockedStatus("Issue with setting up an ingress")
             return
 
-        # 8. Update forward auth relation data with ingressed apps.
+        # Reconcile HPA so Gateway deployment's replicas match the current unit count
+        self._sync_hpa_resources()
```
ca-scribner, line 736: What do you think about folding this into `._sync_gateway_resources()`?  This really is part of managing the Gateway's resources, so logically it feels nice to invoke it there

Also, what about using the gateway reconciler for the HPA too?  I don't think there's a case where we want to add or remove an HPA separate from the Gateway, is there?

tests/integration/helpers.py
```diff
@@ -175,6 +177,26 @@ async def get_route_condition(ops_test: OpsTest, route_name: str) -> Optional[Di
...
+        The HorizontalPodAutoscaler object or None if not found / on error.
+    """
+    model = ops_test.model.name
+    try:
+        c = lightkube.Client()
+
+        return c.get(HorizontalPodAutoscaler, namespace=model, name=hpa_name)
+    except Exception as e:
+        logger.error("Error retrieving HPA %s: %s", hpa_name, e, exc_info=True)
+        return None
```
ca-scribner, line 197: nit: `ops_test` is only used to get `ops_test.model.name`, so just name the arg namespace

```suggestion
async def get_hpa(namespace: str, hpa_name: str) -> Optional[HorizontalPodAutoscaler]:
    """Retrieve the HPA resource so we can inspect .spec and .status directly.

    Args:
        namespace: Namespace of the HPA resource.
        hpa_name: Name of the HPA resource.

    Returns:
        The HorizontalPodAutoscaler object or None if not found / on error.
    """
    try:
        c = lightkube.Client()

        return c.get(HorizontalPodAutoscaler, namespace=namespace, name=hpa_name)
    except Exception as e:
        logger.error("Error retrieving HPA %s: %s", hpa_name, e, exc_info=True)
        return None
```

tests/integration/helpers.py
```diff
@@ -194,6 +216,30 @@ def send_http_request(url: str, headers: Optional[dict] = None) -> bool:
     return resp.status_code == 200
 
 
+def fetch_envoy_peer_metadata_ids(
```
ca-scribner, line 219: nit: We do get the envoy ids as a byproduct here, but "fetch" sounds like we can simply look them up.  I wouldn't see this and think "I need `attempts>>replicas` for this to work".  I can't think of a better name though...

tests/integration/test_charm_ipa.py
```diff
@@ -207,6 +234,31 @@ async def test_route_validity(
...
+            envoy_ids = fetch_envoy_peer_metadata_ids(tester_url, 9)
+            if len(envoy_ids) == expected_count:
+                return envoy_ids
+            await asyncio.sleep(delay)
+        return fetch_envoy_peer_metadata_ids(tester_url, 9)  # Final fetch for assertion
+
+    envoy_ids = await wait_for_envoy_count(1)
+    assert len(envoy_ids) == 1
+
+
```
ca-scribner, line 261: I vote that we don't need to test this.  We confirm in `test_scale` that the current replicas count changed as we expect.  We don't need to confirm that kube-proxy correctly balances load across all pods selected by a Service

tests/scenario/test_gateway.py
```diff
@@ -343,3 +346,69 @@ def _get_listener_given_name(gateway, name: str):
...
+        assert hpa.metadata.namespace == charm.model.name
+
+        spec = hpa.spec
+        assert spec.minReplicas == charm.model.app.planned_units()
+        assert spec.maxReplicas == charm.model.app.planned_units()
+
+        ref = spec.scaleTargetRef
+        assert ref.apiVersion == "apps/v1"
+        assert ref.kind == "Deployment"
+        assert ref.name == "istio-ingress-k8s-istio"
```
ca-scribner, line 370: nit: `charm.model.app.planned_units()` feels a little bit like using an implementation detail as part of the test expectations.  99.9% sure it works as you expect, but imo we should do this:

```suggestion
    n_units = 3
    model_name = "test-model"
    with istio_ingress_context(
        istio_ingress_context.on.update_status(),
        state=scenario.State(leader=True, planned_units=n_units, model=Model(name=model_name)),
    ) as manager:
        charm = manager.charm

        hpa = charm._construct_hpa(n_units)
        assert hpa.metadata.name == istio_ingress_context._app_name
        assert hpa.metadata.namespace == model_name

        spec = hpa.spec
        assert spec.minReplicas == n_units
        assert spec.maxReplicas == n_units

        ref = spec.scaleTargetRef
        assert ref.apiVersion == "apps/v1"
        assert ref.kind == "Deployment"
        assert ref.name == "istio-ingress-k8s-istio"
```

src/charm.py
```diff
@@ -302,12 +322,22 @@ def _handle_auth_config(self, _):
...
         self._sync_all_resources()
 
+    def _on_peers_changed(self, _):
+        """Event handler for whenever peer topology changes."""
+        self._sync_all_resources()
+
     def _on_remove(self, _):
         """Event handler for remove."""
+        if not self.unit.is_leader():
+            return
```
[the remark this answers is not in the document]
ca-scribner: Asked this and a related question [in juju's matrix channel](https://matrix.to/#/!xzmWHtGpPfVCXKivIh:ubuntu.com/$8Gyjn_4KUWzrWflfLnRYadCSCHNzcYfgYt3WqrxDxbc?via=ubuntu.com&via=matrix.org&via=canonical.com)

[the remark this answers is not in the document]
ca-scribner: Some tests using a tester that logs `self.unit.is_leader()` during `remove`... 

### `juju remove-application` when there's a single unit and it is the leader:

Unit removed is still leader at time of handling the `remove` event, 

```
unit-ipa-tester3-0: 15:35:17 WARNING juju.worker.uniter.operation we should run a leader-deposed hook here, but we can't yet
unit-ipa-tester3-0: 15:35:17 INFO juju.worker.uniter.operation ran "stop" hook (via hook dispatching script: dispatch)
unit-ipa-tester3-0: 15:35:17 INFO unit.ipa-tester3/0.juju-log executing _on_remove
unit-ipa-tester3-0: 15:35:17 INFO unit.ipa-tester3/0.juju-log self.unit.is_leader()=True
unit-ipa-tester3-0: 15:35:18 INFO juju.worker.uniter.operation ran "remove" hook (via hook dispatching script: dispatch)
unit-ipa-tester3-0: 15:35:18 INFO juju.worker.uniter unit "ipa-tester3/0" shutting down: agent should be terminated
```

### `juju remove-unit --num-units 1` when there's a single unit and it is the leader:

Unit removed due to scaling is still leader at time of handling the `remove` event, 

```
unit-ipa-tester3-0: 15:25:18 WARNING juju.worker.uniter.operation we should run a leader-deposed hook here, but we can't yet
unit-ipa-tester3-0: 15:25:18 INFO juju.worker.uniter.operation ran "stop" hook (via hook dispatching script: dispatch)
unit-ipa-tester3-0: 15:25:18 INFO unit.ipa-tester3/0.juju-log executing _on_remove
unit-ipa-tester3-0: 15:25:18 INFO unit.ipa-tester3/0.juju-log self.unit.is_leader()=True
unit-ipa-tester3-0: 15:25:19 INFO juju.worker.uniter.operation ran "remove" hook (via hook dispatching script: dispatch)
unit-ipa-tester3-0: 15:25:19 INFO juju.worker.uniter unit "ipa-tester3/0" shutting down: agent should be terminated
```

### `juju remove-unit --num-units 1` when there's two units, and the unit scaled down is not the leader:

There's no mention of the leader-deposed, and the unit is not the leader while handling the remove event

```
unit-ipa-tester3-1: 15:29:11 INFO juju.worker.uniter.operation ran "stop" hook (via hook dispatching script: dispatch)
unit-ipa-tester3-1: 15:29:11 INFO unit.ipa-tester3/1.juju-log executing _on_remove
unit-ipa-tester3-1: 15:29:11 INFO unit.ipa-tester3/1.juju-log self.unit.is_leader()=False
unit-ipa-tester3-1: 15:29:11 INFO juju.worker.uniter.operation ran "remove" hook (via hook dispatching script: dispatch)
unit-ipa-tester3-1: 15:29:11 INFO juju.worker.uniter unit "ipa-tester3/1" shutting down: agent should be terminated
```

From this I think we are **usually** safe, but not safe by design because:
1. if we ever scale away the leader, that unit will remove all our resources
2. if we ever remove during a leader election, there'll be no leader to remove the resources

In (1) we'll have a small downtime, but as soon as a new unit becomes leader it will reconcile the resources back up
In (2) we'll leave resources dangling, but **hopefully** juju will clean them up automatically based on their annotations

ca-scribner (approved):
lgtm, though I have suggestions to refactor a bit

src/charm.py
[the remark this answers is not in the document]
ca-scribner: Since consequences in both events are small, I think this is fine to ignore.  Would matter a lot more if (1) caused more damage

tests/integration/test_charm_ipa.py
```diff
@@ -207,6 +234,31 @@ async def test_route_validity(
...
+@pytest.mark.abort_on_fail
+async def test_gateway_round_robin(ops_test: OpsTest):
+
+    model = ops_test.model.name
+    istio_ingress_address = await get_k8s_service_address(ops_test, "istio-ingress-k8s-istio")
+    tester_url = f"http://{istio_ingress_address}/{model}-{IPA_TESTER}"
+    envoy_ids = fetch_envoy_peer_metadata_ids(tester_url, 9)
+    assert len(envoy_ids) == 3
+
+    await ops_test.model.applications[APP_NAME].scale(scale_change=-2)
```
ca-scribner, line 246: If we don't remove this test, I think we should refactor, maybe adding a fixture or breaking it into two tests.  

The issue I see is that this test changes the cluster state in the middle of the test (scaling the ingress).  That's fine when it works, but if ever debugging the test or trying to keep the environment, you can't run this test twice in a row.  If it has to be that way, that's fine - but probably there's a refactor we could do to make it easier to use

ca-scribner: Easiest way might be a parameterized with an n_ingress_units parameter, so the test sets the units at the start of the test, asserts the expected envoy_count, then resets the scale back to 1 (and resetting the scale to 1 could be through a fixture, that way if the test fails it still gets executed)

src/charm.py
```diff
@@ -182,6 +191,9 @@ def __init__(self, *args):
             self.on[INGRESS_CONFIG_RELATION].relation_broken, self._handle_ingress_config
         )
         self.framework.observe(self.on.leader_elected, self._handle_ingress_config)
+        self.framework.observe(self.on[PEERS_RELATION].relation_joined, self._on_peers_changed)
+        self.framework.observe(self.on[PEERS_RELATION].relation_changed, self._on_peers_changed)
+
```
ca-scribner, line 196: I dont think ['relation_changed'](https://documentation.ubuntu.com/juju/3.6/reference/hook/#endpoint-relation-changed) fires when a unit departs, so I think this is missing scale-down events.  Probably need relation_departed here too

And actually, relation_joined might be unnecessary, but I haven't thought about it too much

tests/integration/test_charm_ipa.py
IbraAoad: HPA tests are now split into `test_hpa_scale_up` and `test_hpa_scale_down`, they aren't a single parameterized test tho and the reason for that is that I want to ensure that when we scale up in the middle of test it doesn't really affect the behavior of the charm in other tests, especially  in `test_route_validity`, and then when we scale down the rest of the tests continue to pass.

Just my way of making sure we didn't break anything without adding more tests and rely on existing ones :D :D, I'm happy to change it to a single parameterized if you don't agree :angel:

src/charm.py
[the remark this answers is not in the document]
IbraAoad: What do you think of doing this:

This would solve 1, but not 2
```
    # if there are still planned units (i.e. a scale-down) OR I'm not leader, skip cleanup
    if self.model.app.planned_units() > 0 or not self.unit.is_leader():
        return
```
and maybe this would actually solve both??

```
    # if there are still units left, skip (this covers scale-downs)
    if self.model.app.planned_units() > 0:
        return
```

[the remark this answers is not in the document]
ca-scribner: hmm good ideas.  With the second proposal, all units will try to remove things but as long as we're tolerant of that (eg: if I try to remove a resource and its not there, I ignore the 404... which I think we do already) then I think that's ok.  I like the second option

src/charm.py
```diff
@@ -316,8 +326,16 @@ def _handle_auth_config(self, _):
         """Event handler for forward_auth config changes."""
         self._sync_all_resources()
 
+    def _on_peers_changed(self, _):
+        """Event handler for whenever peer topology changes."""
+        self._sync_all_resources()
+
     def _on_remove(self, _):
         """Event handler for remove."""
+        # if there are still units left, skip removal
```
ca-scribner, line 335: Suggestion: this feels novel enough that we should describe it in more detail

```suggestion
        """Event handler for remove.

        The objective of this handler is to remove all application-scoped resources when the application is being scaled
        to 0 or removed.  We intentionally do not put this removal action behind a leader guard (eg, behind 
        `if self.unit.is_leader()`) for the reasons discussed 
        [here](https://github.com/canonical/istio-ingress-k8s-operator/issues/16).
        """
        # skip if we are not scaling to 0/removing the application
```

src/charm.py
```diff
@@ -316,8 +326,16 @@ def _handle_auth_config(self, _):
...
+        if self.model.app.planned_units() > 0:
+            return
```
ca-scribner, line 337: Since this is novel and we might want to debug it in future, lets add some helpful logging

```suggestion
        if self.model.app.planned_units() > 0:
            logger.info("Handling remove event: skipping resource removal because application is not scaling to 0.")
            return

        logger.info(
            "Handling remove event: Attempting to remove application resources because application is scaling to 0."
        )

```

src/charm.py
```diff
@@ -894,16 +897,27 @@ def _sync_ext_authz_auth_policy(self, auth_decisions_address: Optional[str]):
...
 
     def _sync_gateway_resources(self):
-        krm = self._get_gateway_resource_manager()
+        unit_count = self.model.app.planned_units()
+
+        # if there are no units left (unit_count < 1), skip reconcile:
+        #  - we assume this is a remove event and the remove hook will clean everything up
+        #  - updating the HPA with zero replicas is invalid
+        if unit_count < 1:
+            return
```
ca-scribner, line 906: iiuc this is guarding against how setting the desired replica count on an HPA to 0 will be rejected by the validating webhook?

I don't see a specific bug with this, but I worry about it not being intuitive.  If unit_count==0, intuitively I'd think this reconciles down to 0 replicas rather than just leaving everything as-is.  Probably you're right that a remove event should occur soon after and all is well, but its hard to guess if there's an edge case here.

Rather than `return`ing here, what if we take `unit_count==0` to mean that the gateway should be removed?  So something like:

```
        krm = self._get_gateway_resource_manager()
        resources_list = []
        
        if unit_count > 0:
            if secret := self._construct_gateway_tls_secret():
                resources_list.append(secret)
                if secret.metadata is None:
                    raise ValueError("Unexpected error: secret.metadata is None")
                tls_secret_name = secret.metadata.name

            resources_list.append(self._construct_gateway(tls_secret_name=tls_secret_name))
            resources_list.append(self._construct_hpa(unit_count))
        krm.reconcile(resources_list)
```

This way if unit_count==0 we do reconcile to 0 rather than ignore the request

tests/integration/test_charm_ipa.py
```diff
@@ -95,10 +97,34 @@ async def test_deployment(ops_test: OpsTest, istio_ingress_charm):
...
+            if hpa.status.currentReplicas == expected_replicas:
+                return True
+            await asyncio.sleep(delay)
+        return False
+
+    assert await wait_for_current_replicas(
+        3
+    ), f"Expected currentReplicas to be 3, got {hpa.status.currentReplicas}"
+
+
```
ca-scribner, line 123: I suggest breaking the scaling tests out into their [own test file](https://github.com/canonical/istio-ingress-k8s-operator/pull/66) rather than linking them with IPA tests.  afaict they're not coupled - if scaling breaks IPA, its because Kubernetes services aren't working as they should.  That feels out of scope for our tests

I wont block on this though

ca-scribner (approved):
lgtm.  a few minor suggestions but generally all good.

src/charm.py
IbraAoad: thaaaanks, will apply your suggestion, alot more efficient imo.

As for the reason, this was added because on application removal we get some events (not sure which one exactly) that fire the `_sync_all_resources` before we reach `on_remove` and that leads to k8s firing a validating webhook cause `planned_untis` is 0.

PR #69, opened 2025-04-28
Append Wildcard Suffix to Unauthenticated Ingress `notPaths` in External Auth AuthorizationPolicy

### LINKED ISSUE #68: Authenticated Ingress List in External Auth `AuthorizationPolicy` Doesn’t Include Subpaths
### Bug Description

When appending the list of authenticated endpoints in the Istio `AuthorizationPolicy` resource for external authorization, the current implementation lists `notPaths` as exact string matches `(e.g., /ingress-auth-hydra)`. However, Istio does not treat these as prefix matches, meaning `/ingress-auth-hydra` does not match `/ingress-auth-hydra/foo`.

To ensure proper exclusion of all subpaths, we need to append wildcard suffixes (/*) to each excluded path during policy generation.

### To Reproduce

1- Juju deploy IAM bundle & oauth2-proxy
2- Juju deploy istio-k8s & istio-ingress-k8s
3- Juju deploy (any workload with nested subpaths) & relate to istio-ingress-k8s:ingress
4- GET {ingress-ip}/{juju-model-name}-{workload-name}/{anysubPath} -> we still get re-routed to oauth2

### Environment

latest/edge

### Relevant log output

```shell
NA
```

### Additional context

_No response_

IbraAoad:
## Issue
Fixes #68

### DIFF
--- a/tests/scenario/test_ingress.py
+++ b/tests/scenario/test_ingress.py
@@ -365,16 +365,15 @@ def test_sync_ingress_resources_with_tls(
 
 
 @pytest.mark.parametrize(
-    "ingress_relations, paths_expected, unauthenticated_paths_expected",
+    "ingress_relations, paths_expected",
     [
         # no relations
-        ([], {}, []),
+        ([], {}),
         # with a single relation that has all data
         (
             [generate_ingress_relation_data("remote-app0", "remote-model0")],
             # (app-name, ingress-relation-name): [list of paths for this app],
             {("remote-app0", "ingress"): ["/remote-model0-remote-app0"]},
-            [],
         ),
         # with multiple related apps on `ingress`
         (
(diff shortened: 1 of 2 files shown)


ca-scribner (approved):
lgtm, but I suggest we test `get_unauthenticated_paths` separately.  Its a simple function and can be tested easily on its own

tests/scenario/test_ingress.py
```diff
@@ -427,12 +431,14 @@ def test_get_routes(
     ) as manager:
         charm: IstioIngressCharm = manager.charm
         routes = charm._get_routes()
+        unauthenticated_paths_actual = get_unauthenticated_paths(routes)
```
ca-scribner, line 434: Lets test this on its own.  It doesn't have anything to do with `_get_routes()`

PR #82, opened 2025-06-03
fix: point integration tests at the current development branch for istio charms

ca-scribner:
## Issue
Previously, integration tests deployed istio-k8s from channel `latest/edge` because that was the leading edge development branch.  Now that we've cut a `1` track for the prior release and a `2` track for current development, we need to point at `2/edge`.  We cannot simply point at `edge`, because `1` is the default track

## Solution
Point the test at `2/edge`

### DIFF
--- a/tests/integration/helpers.py
+++ b/tests/integration/helpers.py
@@ -1,5 +1,6 @@
 import logging
 import ssl
+from dataclasses import dataclass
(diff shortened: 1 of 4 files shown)


lucabello:
Done in https://github.com/canonical/istio-ingress-k8s-operator/pull/145

PR #95, opened 2025-07-25
feat: add terraform module

ca-scribner:
## Solution
Adds a terraform module for the charm.

The README.md for the terraform module is autogenerated using the header from main.tf by [terraform-docs](https://github.com/terraform-docs/terraform-docs) using:

```
sudo snap install terraform-docs
cd terraform
terraform-docs markdown table --output-file README.md . 
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -39,6 +39,7 @@ override.tf.json
 
 # Ignore transient lock info files created by terraform apply
 .terraform.tfstate.lock.info
+*.lock.hcl
 
 # Ignore CLI configuration files
 .terraformrc


terraform/main.tf
```diff
@@ -0,0 +1,39 @@
...
+ *   channel = "1/stable"
+ * }
+ * ```
+ *
+ * ```sh
+ * $ terraform apply
+ * ```
+ */
+
+resource "juju_application" "this_app" {
```
adhityaravi, line 25: `this_app` should be renamed to an appropriate resource name. istio_ingress or sth.

adhityaravi (approved):
LGTM with a couple of comments. 

While we can debate if adding tests to the tf module can be useful or not, I would like for us to have atleast a simple example tf file that we can run to validate the working of the tf module for sanity. 

Example from istio-beacon - https://github.com/canonical/istio-beacon-k8s-operator/blob/main/terraform/tests/examples/istio-beacon.tf

terraform/main.tf
ca-scribner: I don't care strongly about this and can change it.  

afaiu using an appropriate name:
* doesn't have a positive functional effect (this name isn't exposed anywhere, like when using this in a solution-level module)
* has a minor negative effect (it adds the chore of [keeping this line up to date](https://github.com/canonical/istio-ingress-k8s-operator/pull/95/files#diff-39c11374cc25633f69e49ef1485fa8d4dfd4c33cde3027c8c20a2b84346ebf7cR2))

Am I wrong about this?  I don't object to writing something meaningful if it matters, but there should be a reason beyond "it looks nice when we're writing the module"

terraform/version.tf
```diff
@@ -0,0 +1,9 @@
+terraform {
+  required_version = ">= 1.5"
+  required_providers {
+    juju = {
+      source  = "juju/juju"
+      version = ">= 0.14.0"
+    }
+  }
+}
```
ca-scribner, line 9: Responding to [this comment](https://github.com/canonical/istio-ingress-k8s-operator/pull/95#pullrequestreview-3061366040) somewhere meaningless because I wanted a thread for discussion :)  This has to do with what we include in `/tests`, not anything to do with the version.tf file

This is mainly my own ignorance.  What's the difference in outcomes between running:
* [this example from istio-beacon](https://github.com/canonical/istio-beacon-k8s-operator/blob/main/terraform/tests/examples/istio-beacon.tf)
* `terraform init; juju add-model test; terraform apply -var model=test -var channel="2/edge --auto-accept" -out plan` in the `/terraform` directory?

I get that we also deploy istio-k8s, though the example doesn't actually depend on it for the outcome (the status of istio-beacon at the end doesn't matter).  So the example kinda feels like a longer version of a bash oneliner.

The [istio-beacon](https://github.com/canonical/istio-beacon-k8s-operator/blob/main/terraform/tests/examples/istio-beacon.tf) example isn't so crazy since its simple, but with Kiali we can either:
* do a trivial example where we just deploy kiali and let it go to blocked because dependencies are missing
* deploy prometheus, istio, grafana, and then relate them up, and then we've got a full integration test

Maybe the second option is what we need to do, but it feels overlapping with our integration tests...

terraform/variables.tf
```diff
@@ -0,0 +1,45 @@
...
+variable "config" {
+  description = "Map of the charm configuration options"
+  type        = map(string)
+  default     = {}
+}
+
+variable "constraints" {
+  description = "String listing constraints for this application"
+  type        = string
+  default     = ""
```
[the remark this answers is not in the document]
ca-scribner: If we do keep the empty string default, we also need to bump the juju terraform provider to the version that supports this fix

terraform/main.tf
adhityaravi: Haha it does look nice when we're writing the module :P

But i think youre right, the resource identifier is used only by terraform for handling addresses in state files but even then it is module scoped, so we will not have any problem using the same resource id across all our charms. The caveat being we continue to use a tf module per charm.

When we will have to have more than one resource in a tf module we would need to use a unique id per resource. 

So, i leave it up to you to use "this_app" or something nicer.

ca-scribner: Yes, I agree that a *solution level* module would need something better than this.  For a single-charm module, I think it is a throwaway statement

I'm also influenced by minimal maintenance and making this as templatable as possible.  If maintenance goes up a little or templatability goes down, I'd like to see a return on the investment.

That being said, this isn't a hill I'd die on :P

terraform/variables.tf
dstathis, line 21:
```suggestion
  default     = "arch=amd64"
```

terraform/main.tf
dstathis: Resolving in the interest of getting things done. Feel free to continue the discussion.

terraform/variables.tf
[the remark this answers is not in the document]
dstathis: Fixed.

PR #112, opened 2025-10-27
feat: supports multiple ports and protocols

### LINKED ISSUE #107: Add istio-tcp-route or istio-ingress-route relation.
### Enhancement Proposal

Tempo needs some features that are not available in the regular ingress relation.
- tcp routes
- multiple routes

Therefore we will need to provide a more powerful relation.

adhityaravi:
## Issue
fixes #107

This PR is a little involved that adds support to multiple ports and protocols in a clean way.

### Context
The ingress needs to be able to open multiple ports (not just the std. 80 and 443) when requested via a different, advanced charm relation interface.

The good - the K8s gateway listener does not differentiate between HTTP and gRPC (unlike for eg., traefik which has a special `h2c` context for `http/2 cleartext` protocol). This simplifies the opening multiple ports (listeners) logic. And HTTPRoutes and GRPCRoutes can attach to the same listener.

The bad - The current charm code is not flexible enough to easily accommodate multiple listeners, rather it works purely based the relation data structure from the `ingress v2` library.

The ugly - According to the istio [docs](https://istio.io/latest/docs/ops/configuration/traffic-management/protocol-selection/#http-gateway-protocol-selection), the gateway is unable to determine the protocol of the backend unless explicitly defined using the portname or the appProtocol. By default it wll use http/1.1. This is problematic because for gRPC backends, the gateway uses http/1.1 as well by default (because there is no valid port-name nor appProtocol). Hence gRPC routing fails. The charm has no control neither over the port name (auto generated by Juju) nor over the appProtocol (Juju controls the service so we cant reliably patch this).

### Solution overview
The idea behind the solution to support multiple ports and protocols (HTTP and gRPC now, TCP/UDP later if needed?) is

1. Obviously add new relations for advanced route requesting
2. Provide a charm library for assisting in the creation of Gateway listeners and routes (unlike the `traefik_route.py` lib we wont be raw dogging with plain yaml configs rather provide helper dataclasses to construct listeners and routes following the `istio-beacon` pattern)
3. Aggregate the standard listeners and the charm requested listeners
4. Aggregate the routes requested from the `ingress` and `istio-ingress-route` relations, remove any duplicates, punish the duplicate requests by not providing the ingress to conflicted apps (legacy behavior)
5. For every gRPC backend, reconcile a `DestinationRule` resource with [useClientProtocol](https://istio.io/latest/docs/reference/config/networking/destination-rule/#ConnectionPoolSettings-HTTPSettings-use_client_protocol) set to `True`. This is a workaround for the gateway not being able to detect the backends protocol. This instructs the gateway to use the same protocol as that of the client request. **This creates additional overheads. But there seemingly is no other way at the moment.**

### Code structure change
To support the above code  flow tastefully there was a change in code structure needed

The old pattern was to construct the K8s Gateway resources (Gateway, HTTPRoutes etc) directly from the `ingress` relation data. On adding a new relation interface for ingress, this means that we need 2x the code for the same logic (a method that converts `ingress` relation data to HTTPRoute resource, a method that converts `instio-ingress-route` relation data to a HTTPResource etc.)

Hence adapter schemas were used to aggregate all the listeners and routes from both the ingress realtion interfaces together into a standard datatype. And the listeners and routes are deduplicated, and reconciled together. This makes the charm easily extensible to other protocols that aren't currently supported by this PR (TCP, UDP). Also upgrading to major versions of the `ingress` charm library will be easy in the future (since we only need to care about changing the adapter methods and not worry about the charm logic itself).

### Drive-bys
The above also meant that the `charm.py` became +2000 line long. For sake of sanity and code maintainability, the utility code is abstracted into the `utils.py` file.

The `lightkube-extensions` package was updated to use the [`client.patch`](https://lightkube.readthedocs.io/en/latest/reference/client/#lightkube.Client.patch) with flexible [`PatchType`](https://lightkube.readthedocs.io/en/latest/reference/types/#lightkube.types.PatchType) instead of hardcoded usage of `client.apply`. This was needed because the `client.apply` doesnt remove the stale fields (listeners, hostname etc) when we reconcile a resource with the same name again. This also completes this [TODO](https://github.com/canonical/istio-ingress-k8s-operator/blob/f86d9b0887a3fb675ccacf82b37acf1afe647c27/src/charm.py#L956) and hence the `_remove_hostname_if_present` was removed.

### Is the change breaking?
No. This PR isn't meant to have any breaking changes. The original `ingress` relation should continue to work as it is.

## Tandem PRs
- [lightkube-extensions](https://github.com/canonical/lightkube-extensions/pull/8)

## Testing Instructions
Unit and integration tests are added. A `tester-grpc` charm was added to verify gRPC routing. the HTTP and the gRPC tester charms can be used to verify the ingress working.

### DIFF
--- a/lib/charms/istio_ingress_k8s/v0/istio_ingress_route.py
+++ b/lib/charms/istio_ingress_k8s/v0/istio_ingress_route.py
@@ -195,7 +195,7 @@ class ProtocolType(str, Enum):
         >>> # Consumer specifies GRPC
         >>> Listener(name="grpc", port=9090, protocol=ProtocolType.GRPC)
         # If TLS available: Gateway serves HTTPS (HTTP/2 + TLS) on port 9090
-        # If TLS not available: Gateway serves HTTP (h2c) on port 9090
+        # If TLS not available: Gateway serves HTTP on port 9090
     """
 
     HTTP = "HTTP"
@@ -211,7 +211,7 @@ def to_gateway_protocol(protocol: ProtocolType, tls_enabled: bool = False) -> st
     """Map application protocol to Gateway API protocol.
 
     The Gateway API doesn't have separate HTTP/gRPC protocol types.
-    Both use HTTP/2, with the difference being in the route type (HTTPRoute vs GRPCRoute).
+    Both use HTTP, with the difference being in the route type (HTTPRoute vs GRPCRoute).
 
     Args:
         protocol: Application-level protocol (HTTP or GRPC)
--- a/src/charm.py
+++ b/src/charm.py
@@ -728,14 +728,16 @@ class IstioIngressCharm(CharmBase):
         * Check authentication configuration.
         * Publish or clear the auth_decisions_address in ingress-config, if related.
         * If auth relation exists but no decisions address, set to blocked and remove gateway.
-        * Fetch route information from the ingress relation
+        * Fetch route information from the ingress relation.
+        * Fetch route and listeners information from the istio-ingress-route relation.
+        * Aggregate and deduplicate the routes and listeners from all supported ingress relations.
         * Synchronize external authorization configuration.
             - If missing valid ingress-config relation when auth is provided, set to blocked and remove gateway.
         * Reconcile HPA and gateway resources to align replicas with unit count and ensure gateway readiness.
         * Validate the external hostname.
-        * Synchronize ingress resources
+        * Synchronize ingress resources.
         * Publish route information to ingressed applications
-        * Set up the proxy service
+        * Set up the proxy service.
         * Update forward auth relation data with ingressed apps.
         * Request certificate inspection.
         """


lib/charms/istio_ingress_k8s/v0/istio_ingress_route.py
```diff
@@ -0,0 +1,717 @@
...
+
+To get started using the library, you just need to fetch the library using `charmcraft`.
+
+```shell
+cd some-charm
+charmcraft fetch-lib charms.istio_ingress_k8s.v0.istio_ingress_route
+```
+
+To use the library from the provider side (istio-ingress):
+
```
dstathis: I think we can leave out the instructions for the provider side since it is only expected to be used by exactly one charm. Or at the very least, we should put the requirer docs first, as they are much more useful.

lib/charms/istio_ingress_k8s/v0/istio_ingress_route.py
```diff
@@ -0,0 +1,717 @@
...
+        # If TLS available: Gateway serves HTTPS (HTTP/2 + TLS) on port 9090
+        # If TLS not available: Gateway serves HTTP (h2c) on port 9090
+    """
+
+    HTTP = "HTTP"
+    GRPC = "GRPC"
+
+    # TODO: Extend support to L4 protocols?
+    # TCP = "TCP"
+    # UDP = "UDP"
```
dstathis: Open an issue instead of leaving this comment.

dstathis (approved):

lib/charms/istio_ingress_k8s/v0/istio_ingress_route.py
adhityaravi: done

Issue #117, opened 2025-11-17
Consolidate all data models into the istio_ingress_route library


adhityaravi:
### Enhancement Proposal

Currently the data models used in the istio ingress charm are distributed between the `models.py`, `utils.py` and in the `istio_ingress_route.py` library. This was intially done to introduce and validate the istio ingress route relation without major refactoring to the existing code. 

But with growing need for new features to the istio ingress route has led to some duplication, bad code clarity and maintenance headaches. So, we need to consolidate and deduplicate all the data models into a single source of truth - `istio_ingress_route.py` library.

why the library? because the models will also be used by charm authors to build custom ingress routes. 

Eventually - the libraries must be migrated into [charmed-service-mesh-helpers](https://github.com/canonical/charmed-service-mesh-helpers) or another py package. But this can be done when we start to provide the ingress library itself as a namespaced python package. (this way we dont need to document all the custom dependencies we use in the library)

dstathis:
I don't think we should put the models in istio-ingress-route, but rather, move everything into the [charmlibs](https://github.com/canonical/charmlibs) repo under a mesh directory so developers would import the code like this:


```
from charmlibs.mesh.relations import ingress_route_requirer
from charmlibs.mesh.data_models import randomDataModel
```

dstathis:
This should include adding a schema.py for each of our interfaces and also should include migrating our charms to use this library.

dstathis:
This should include the service_mesh library as well.

adhityaravi:
This issue can be closed. We have addressed this using `charmlibs` and `canonical-service-mesh` libraries

PR #120, opened 2025-11-18
feat: istio_ingress_route supports route filters

### LINKED ISSUE #116: istio-ingress-route needs to support filter matches
### Enhancement Proposal

KF team needs the following HTTPRoute resource (check the comment). currently the istio ingress route libraries HTTPRoute model only allows a path match and no filter matches. So we need to extend this to support the filter matches as well.

adhityaravi:
## Issue
fixes #116 

## Solution decisions
1. Some data models are migrated from the `models.py` into the `istio_ingress_route.py` library. This is temporary to unblock this issue. The longer plan is to consolidate the data models into charmlibs. See #117 
2. Currently only the `URLRewriteFilter` and the `RequestRedirectFilters` are included as supported filters. But an issue is added to eventually extend support to all the K8s gateway filters. See #119

## Testing
1. The `RequestRedirectFilter` is alread used for HTTP to HTTPS redirect when tls is enabled. so the tls unit and integrations tests covers this one.
2. The `URLRewriteFilter` is explicitly used and tested in the `tester-http` charm.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -566,61 +566,6 @@ class IstioIngressCharm(CharmBase):
             spec=gateway.spec.model_dump(exclude_none=True),
         )
 
-    # def _construct_httproute(
-    #     self,
-    #     service_name: str,
-    #     namespace: str,
-    #     port: int,
-    #     strip_prefix: bool,
-    #     prefix: str,
-    #     section_name: str,
-    # ):
-    #     http_route = HTTPRouteResource(
-    #         metadata=Metadata(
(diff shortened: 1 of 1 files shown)


src/models.py
```diff
@@ -92,49 +94,6 @@ class HTTPRouteMatch(BaseModel):
     path: HTTPPathMatch
 
 
-class PrefixPathConfig(BaseModel):
```
adhityaravi: Route filter related models are for now migrated to the istio_ingress_route library. hence removed from here.

PR #151, opened 2026-03-30
feat: adds support to upstream ingress

adhityaravi:
## Issue
Fixes #142 

## Solution
Adds an `upstream-ingress` relation over the `ingress` interface. to access apps in a chained ingress setup

## Testing Instructions
Can be tested by deploying a `traefik-k8s` charm infront of the istio-ingress-k8s charm and trying to reach the workloads via the outermost ingress

### DIFF
--- a/charmcraft.yaml
+++ b/charmcraft.yaml
@@ -123,6 +123,14 @@ requires:
   forward-auth:
     interface: forward_auth
     limit: 1
+  upstream-ingress:
+    interface: ingress
+    limit: 1
+    description: |
+      Configure to use an ingress upstream of this gateway. Useful when you want an internal
+      gateway that is then exposed by an additional external ingress.
+      When upstream ingress is configured, the URLs published to downstream consumers will use
+      the upstream ingress's address rather than this gateway's load balancer address.
 
 resources:
   metrics-proxy-image:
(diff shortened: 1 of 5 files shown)


adhityaravi:
@dstathis @simskij The chaining works fine when the upstream charm is a non Gateway API charm like the `traefik-k8s` but doesnt work when the upstream ingress is a Gateway API charm like the `istio-ingress-k8s`. Because

1. the ingress v2 library does not provide a way to specify a custom service name. For gateway API, the juju created service is useless. The actual LB service is named differently and there is no way to forward this information at this moment. 
2. Even if that were possible (that seems relatively simple), which we shouldnt use for chaining ingresses anyway. when the downstream ingress has an external hostame configured or has an external ip, we dont handle this yet in the `istio-ingress-k8s` charm because of this [issue](https://github.com/canonical/istio-ingress-k8s-operator/issues/96) - ironically we thought this wont come up anytime soon but it is here sooner.

Not sure how other gateway charms like the [gateway-api-integrator](https://github.com/canonical/gateway-api-integrator-operator#) are going to handle this. But do you guys think we should reopen the issue from Andrew and handle this in istio ingress?

dstathis (approved):
Looks good. Very clean!

PR #155, opened 2026-04-16
fix: add all endpoints to terraform output

gycsaba96:
## Issue

Terraform output was incomplete. See #140.

## Solution

Add the missing endpoints.


dstathis (approved):
LGTM!

PR #157, opened 2026-04-24
fix: enter blocked status when the same ingress path is requested by multiple apps

### LINKED ISSUE #57: Set charm to BlockedStatus when multiple charms request the same route through an ingress
### Enhancement Proposal

Currently, if multiple charms request the same path through (eg: `INGRESS/some-path`) the ingress we:
* give nobody that route
* log an error message to the charm logs

Notably, we **do not** set the charm to BlockedStatus even though this is a misconfiguration that requires user intervention.  This is because we don't want the charm to halt when this happens (because other ingresses should be maintained), but the charm isn't written to make this easy.  

The goal of this issue is to refactor the charm to:
* capture a BlockedStatus when path collision happens
* continue with the rest of the charm logic
A possible solution would be something like what we do with [kiali's `reconcile`](https://github.com/canonical/kiali-k8s-operator/blob/507eb7e6485d8181cabb225ffb56fcd9b1995d9e/src/charm.py#L143)).

gycsaba96:
## Issue

When multiple apps requested the same ingress path, the charm only communicated the problem through log messages. A blocking status is more visible. Fixes #57.

## Solution

A list of "soft blockers" is collected during the synchronization of all resources. This final status is only set at the end to avoid breaking other functionality.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -727,6 +727,7 @@ class IstioIngressCharm(CharmBase):
         """Synchronize all resources including authentication, gateway, ingress, and certificates.
 
         Flow:
+        * Create a list of events (soft blockers) that do not break the flow but set a blocked state at the end.
         * Check authentication configuration.
         * Publish or clear the auth_decisions_address in ingress-config, if related.
         * If auth relation exists but no decisions address, set to blocked and remove gateway.
@@ -742,10 +743,14 @@ class IstioIngressCharm(CharmBase):
         * Set up the proxy service.
         * Update forward auth relation data with ingressed apps.
         * Request certificate inspection.
+        * Set the final status based on soft blockers.
         """
         if not self.unit.is_leader():
             self.unit.status = ActiveStatus("Backup unit; standing by for leader takeover")
             return
+
+        # Create a list of events (soft blockers) that do not break the flow but set a blocked state at the end.
+        soft_blockers = []
 
         # Check authentication configuration.
         auth_decisions_address = self._get_oauth_decisions_address()
@@ -793,9 +798,10 @@ class IstioIngressCharm(CharmBase):
         valid_grpc_routes, grpc_apps_to_clear = deduplicate_grpc_routes(istio_grpc_routes)  # ipa relation doesnt support grpc routes.
 
         # Clear conflicts from original structures (needed for publishing back to apps)
-        # TODO: Capture a BlockedStatus here if there's duplicates
-        #  (https://github.com/canonical/istio-ingress-k8s-operator/issues/57)
         all_apps_to_clear = http_apps_to_clear | grpc_apps_to_clear
+        logger.error(str(all_apps_to_clear))
+        if len(all_apps_to_clear) > 0:
+            soft_blockers.append("Route conflict detected. Check the logs for more information.")
         clear_conflicting_routes(
             application_route_data, istio_ingress_route_configs, all_apps_to_clear
         )
@@ -877,6 +883,13 @@ class IstioIngressCharm(CharmBase):
             "Requesting CertHandler inspect certs to decide if our CSR has changed and we should re-request"
         )
         self.on.refresh_certs.emit()
+
+        # Set the final status based on soft blockers
+        self._set_final_status(soft_blockers)
+
+    def _set_final_status(self,soft_blockers):
+        if len(soft_blockers) > 0:
+            self.unit.status = BlockedStatus(str(soft_blockers[0]))
 
     def _get_oauth_decisions_address(self) -> Optional[str]:
         """Retrieve the auth configuration decisions_address if it exists.
--- a/src/charm.py
+++ b/src/charm.py
@@ -8,7 +8,7 @@ import ipaddress
 import logging
 import re
 import time
-from typing import Any, Dict, List, Optional, cast
+from typing import Any, Dict, List, Optional, Tuple, cast
 from urllib.parse import urlparse
 
 from charmed_service_mesh_helpers import (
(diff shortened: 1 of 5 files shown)


adhityaravi (changes requested):
Good job on the first istio PR!

Considering this [issue comment](https://github.com/canonical/istio-ingress-k8s-operator/issues/57#issuecomment-3655947947), we should rather add a status handler instead of adding a custom list and raising the first blocked status.

We should let Ops do this. Ops can automatically collect all the available statuses and raise the worst one on reconciliation. Example implementation for reference

- https://github.com/canonical/tempo-operators/blob/493bec3c0cc9780ea59e1f28c587d47324ab3a32/coordinator/src/charm.py#L504
- https://github.com/charmarr/charmarr/blob/1d51dda73553e79cbae1505a9fb28b552f3ff3c7/charms/gluetun-k8s/src/charm.py#L433

src/charm.py
```diff
@@ -156,8 +156,19 @@
 class IstioIngressCharm(CharmBase):
     """Charm the service."""
 
+    _stored = StoredState()
```
adhityaravi: Adding StoredState in general is  not a good idea for holistic charms (the charms using the reconciliation pattern. all of our charms use this pattern). It makes the charms kinda "stateful" which is an anti-pattern for the reconciliation pattern.

src/charm.py
```diff
@@ -535,13 +546,15 @@ def _get_lb_external_address(self) -> Optional[str]:
 
     def _is_ready(self) -> bool:
```
adhityaravi, line 611: Since we are adding a status handler, the `_is_ready` method should not add any status. It should purely tell if the charm and its workload is in a ready state or not. The status evaluation logic can move into the status handler method.

src/charm.py
```diff
@@ -878,6 +904,12 @@ def _sync_all_resources(self):
         )
         self.on.refresh_certs.emit()
 
+    def _on_collect_status(self, event: CollectStatusEvent):
+        if self._stored.are_routes_removed:
```
adhityaravi: there is no need for the `_stored` object here. You can pack the entire status evaluation logic here. For ex.

```python
def _on_collect_status(self, event: CollectStatusEvent):
    if not self.unit.is_leader():
        event.add_status(
            (ActiveStatus("Backup unit; standing by for leader takeover")
        )
        return  # add a return here because the non-leader units needn't process other status information. 

    if not self._is_deployment_ready():
        event.add_status(
            BlockedStatus("Gateway k8s deployment not ready, is istio properly installed?")
        )
        # no return for other status evals.
```

if you want it to be nicer looking you can break down this single status handler block into smaller blocks. one method per status eval. Something like

```python
def _on_collect_status(self, event: CollectStatusEvent):
    self._collect_leader_status()
    if self.unit.is_leader():
        self._collect_deployment_status()
        self......

def _collect_leader_status():
    if not self.unit.is_leader():
        event.add_status(
            (ActiveStatus("Backup unit; standing by for leader takeover")
        )

def _collect_deployment_status():
    if not self._is_deployment_ready():
        event.add_status(
            BlockedStatus("Gateway k8s deployment not ready, is istio properly installed?")
        )
```

src/charm.py
```diff
@@ -878,6 +904,12 @@ def _sync_all_resources(self):
         )
         self.on.refresh_certs.emit()
 
+    def _on_collect_status(self, event: CollectStatusEvent):
```
adhityaravi, line 1089: Probably the existing private methods dont have a doc string. But its always nice to have an one line doc string telling what this method is doing

adhityaravi (changes requested):
Most of our and the o11y charms use the reconciliation pattern. Basically we don't care about the state and let the reconciler drive to the desired state. So the charms are truly stateless. I have added some comments so we adhere to this pattern.

src/charm.py
gycsaba96: Done.

src/charm.py
gycsaba96: Done.

src/charm.py
gycsaba96: Done.

src/charm.py
gycsaba96: Done.

src/charm.py
```diff
@@ -1097,6 +1086,81 @@ def _sync_all_resources(self):
...
+        """
+        status_collectors = [
+           self._collect_leadership_status,
+           self._collect_auth_decisions_status,
+           self._collect_route_collision_status,
+           self._collect_external_authorization_status,
+           self._collect_readiness_status,
+           self._collect_external_hostname_status,
+        ]
+        for collector in status_collectors:
```
adhityaravi, line 1103: This is not a blocker, but if we add the status evaluators into a list and exit early, we are essentially deciding which status comes first. And if we dont consciously order the statuses, we wont be able to raise the worst status fast ie we could be stuck on a Maintenance status for some time before a Blocked status is raised. 

I agree we would save a few function evals if we exit early, but I dont think we gain much by saving that time. I think we should simply add all the status events and let Ops pick the one for us. this is how the other observability charms do it. But I am not too opinionated. You can be the judge of what you want to do.

adhityaravi (approved):
awesome! the code already looks much better than what we have in main. 

I just have one comment, but feel free to be the the judge of if you want to address it or not.

adhityaravi (commented):
(Also please add a resolution keyword in your PR description like "Fixes #57" so the issue is closed when the PR is merged)

src/charm.py
gycsaba96: I was also thinking a lot about this. However, it is not just about saving time (although I have seen sleep commands). Some checks doesn't make sense or potentially lead to an exception without ensuring their preconditions (e.g., when the unit is not a leader, or checks depend on each other.)

Alternatively, we could always re-check the necessary conditions in each collect function. However, that felt very redundant. Moreover, the current solution still allows for collecting everything: the future contributor writes the collector with every suitable check (without returning anything), and places the function to the beginning of the list.

> [...] we could be stuck on a Maintenance status for some time before a Blocked status is raised.

This is true, but it is an option and not a limitation. If a check doesn't want to depend on the readiness, it can be placed earlier or to the beginning of the list.

adhityaravi: Actually now that you mention it, I think except for leader check, all the other status methods should be behind a leader guard. This is currently kinda implicit because the leader check is placed at the top of the list. At the least we should add a comment saying that the status checks should be done by the leader unit.

adhityaravi: But the rest looks good to me!

gycsaba96: Done.

PR #161, opened 2026-04-30
fix: includes all requested port listeners in the L4 policy

### LINKED ISSUE #160: The ingress L4 policies are overwritten incase of multiple ports on the consumer
### Bug Description

Istio ingress charm creates an L4 policy to allow itself to communicate with the consumer. but when the consumer charm is having multiple ports open via the `istio-ingress-route` then the ingress charm is creating the "same" policy with different ports which leads to the policies overwriting each other.

### To Reproduce

1. deploy tempo
2. deploy istio (core + ingress)
4. deploy istio-beacon
5. integrate mesh and ingress in tempo
6. check the ingress L4 policy

### Environment

`istio-ingress-k8s` dev/edge (track/2 should have the same issue)

### Relevant log output

```shell
-
```

### Additional context

-

adhityaravi:
## Issue
Fixes #160 

## Solution
Include all requested port listeners requested by the downstream application in the L4 policy constructed in the charm

## Context
The `istio-ingress-route` interface allowed the downstream application request more than one port listener. And the L4 policy creation was logic was only covering the case where we had a single fix listener port. This PR addresses this gap.

## Testing Instructions
Follow the itests for the `istio-ingress-route` interface or test with the [tempo](https://github.com/canonical/tempo-operators) charm

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -691,7 +691,7 @@ class IstioIngressCharm(CharmBase):
         )
 
     def _construct_auth_policy_from_ingress_to_target(
-        self, target_name: str, target_namespace: str, target_port: int
+        self, target_name: str, target_namespace: str, target_ports: List[int]
     ):
         """Return an AuthorizationPolicy that allows the ingress workload to communicate with the target workload."""
         return AuthorizationPolicy(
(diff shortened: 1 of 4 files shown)


dstathis (approved):

dstathis:
LGTM

PR #163, opened 2026-05-04
fix: extends default include headers with all standard headers

adhityaravi:
## Issue
The standard headers included for forward authentication does not include all the standard headers that would be required by the proxy auth. This PR extends the default with required default headers for oauth2-proxy to be able to work with istio-ingress-k8s when istio ingress is behind an outer ingress.


dstathis (approved):

adhityaravi:
Merging with red CI since the charmhub api is bork at the moment

PR #165, opened 2026-05-06
fix: improves regressed status collector performance

adhityaravi:
## Issue
The recently added implementation of the status handler code in https://github.com/canonical/istio-ingress-k8s-operator/pull/157 is causing the status updates to lag and sometimes preventing the charm from even going into an active state because of the following reasons

- the reconciler and the status collector both are polling the gateway readiness for 10 mins. meaning we are stuck in maintenance for almost 20 mins even before getting an update
- since the reconciler and the status hook is stuck, it prevents juju from retrying the hooks as well
- the custom for-loop instead of the Ops native way is interfering with how Ops is updating the status which on top basically puts the entire charm in a hanged state
- the above conditions worsen when there are multiple ingresses and any delays in isitos attempts to deploy the gateway are exponentially worsened. 

While eventually the charms reconcile and the status is turned active, 20+ minutes for the ingress charm to maybe settle is bad UX

## Solution
This fix
- sticks to the Ops native way for adding all the status events and let Ops pick the right one instead of a custom for loop
- makes the status event return early instead of polling for gateway readiness

## Drive by
fixes failing itest to adapt to the new lib changes to include more standard request headers

### DIFF
--- a/lib/charms/istio_k8s/v0/istio_ingress_config.py
+++ b/lib/charms/istio_k8s/v0/istio_ingress_config.py
@@ -115,7 +115,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 3
+LIBPATCH = 4
 
 DEFAULT_RELATION_NAME = "istio-ingress-config"
 
--- a/src/charm.py
+++ b/src/charm.py
@@ -550,28 +550,31 @@ class IstioIngressCharm(CharmBase):
         kgm.delete()
 
     def _is_deployment_ready(self) -> bool:
-        """Check if the deployment is ready after 10 attempts."""
+        """Check if the deployment is ready after multiple attempts."""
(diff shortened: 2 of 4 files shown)


lib/charms/istio_k8s/v0/istio_ingress_config.py
```diff
@@ -115,7 +115,7 @@ def _dump_data(data: Dict) -> Dict:
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 3
+LIBPATCH = 4
```
dstathis, line 118: Zero changes library patch? Do we know why this is a thing?

dstathis (approved):

lib/charms/istio_k8s/v0/istio_ingress_config.py
adhityaravi: yea, because i accidentally manually added the changes without actually using charmcraft fetch previously

PR #181, opened 2026-06-05
chore: utils.py refactoring

### LINKED ISSUE #118: Normalization utilities (`utils.py`) should not need intermediate data classes
### Enhancement Proposal

Currently the `ipa` and `istio_ingress_route` relation data are converted into a standard data model in the Utils and then they are converted into `HTTPRouteResource` etc

But once all the data models are standardized in a single source of truth #117, this intermediary data models are just bloat. we can directly normalize the relation data into K8s Resource model.

gycsaba96:
Closes #118 

This PR aims to use the final k8s objects during normalization and deduplication. `GatewayListener` was removable because it was a flat version of the k8s object. However, `HTTPRoute` and `GRPCRoute` contained more information than their k8s counterpart. Thus, they cannot be fully eliminated. Moreover, `GRPCRoute` and `HTTPRoute` introduced some `route.something.somethingelse.thisiswahtineed` patterns. 

(I am OK with keeping only the first two commits. Then, `charm.py` can be simplified by extracting the creation of those nested k8s `...Resource` objects into their own functions.)

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -27,7 +27,6 @@ from canonical_service_mesh.k8s.resource_manager import (
 from canonical_service_mesh.k8s.types.istio import AuthorizationPolicy
 from canonical_service_mesh.models import (
     AllowedRoutes,
-    GatewayTLSConfig,
     GRPCRouteResource,
     GRPCRouteResourceSpec,
     GRPCRouteRule,
@@ -39,7 +38,6 @@ from canonical_service_mesh.models import (
     Listener,
     Metadata,
     ParentRef,
-    SecretObjectReference,
 )
 from canonical_service_mesh.models.istio import (
     AuthorizationPolicySpec,
--- a/src/charm.py
+++ b/src/charm.py
@@ -27,9 +27,6 @@ from canonical_service_mesh.k8s.resource_manager import (
 from canonical_service_mesh.k8s.types.istio import AuthorizationPolicy
 from canonical_service_mesh.models import (
     AllowedRoutes,
-    GRPCRouteResource,
-    GRPCRouteResourceSpec,
-    GRPCRouteRule,
     IstioGatewayResource,
     IstioGatewaySpec,
     Listener,
@@ -1413,31 +1410,11 @@ class IstioIngressCharm(CharmBase):
 
         for route in grpc_routes:
             # Derive listener name from Gateway protocol and port
-            listener_name = f"{route['listener_protocol'].lower()}-{route['listener_port']}"
+            listener_name = f"{route.listener_protocol.lower()}-{route.listener_port}"
 
             # Construct GRPCRoute resource from normalized data
-            grpc_route_resource = GRPCRouteResource(
-                metadata=Metadata(
-                    name=route["name"],
-                    namespace=route["namespace"],
-                ),
-                spec=GRPCRouteResourceSpec(
-                    parentRefs=[
-                        ParentRef(
-                            name=self.app.name,
-                            namespace=self.model.name,
-                            sectionName=listener_name,
-                        )
-                    ],
-                    rules=[
-                        GRPCRouteRule(
-                            matches=route["matches"],  # Already charm GRPCRouteMatch models
-                            backendRefs=route["backend_refs"],  # Already charm BackendRef models
-                            filters=route["filters"] if route["filters"] else None,
-                        )
-                    ],
-                ),
-            )
+            grpc_route_resource = route.resource
+            grpc_route_resource.spec.parentRefs = [ParentRef(name=self.app.name,namespace=self.model.name,sectionName=listener_name)]
 
             # Convert to lightkube resource
             grpcroute_lk_resource = RESOURCE_TYPES["GRPCRoute"]
@@ -1476,7 +1453,9 @@ class IstioIngressCharm(CharmBase):
                 backend_ports.setdefault(key, set()).add(backend_ref.port)
 
         for route in grpc_routes:
-            for backend_ref in route["backend_refs"]:
+            backend_refs = route.resource.spec.rules[0].backendRefs
+            assert backend_refs is not None
+            for backend_ref in backend_refs:
                 key = (backend_ref.name, backend_ref.namespace)
                 backend_ports.setdefault(key, set()).add(backend_ref.port)
 
@@ -1505,7 +1484,9 @@ class IstioIngressCharm(CharmBase):
         seen_backends = set()
 
         for route in grpc_routes:
-            for backend_ref in route["backend_refs"]:
+            backend_refs = route.resource.spec.rules[0].backendRefs
+            assert backend_refs is not None
+            for backend_ref in backend_refs:
                 # One DR per unique (service, namespace) - not per port
                 backend_key = (backend_ref.name, backend_ref.namespace)
                 if backend_key not in seen_backends:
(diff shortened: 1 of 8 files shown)


src/charm.py
```diff
@@ -641,23 +632,14 @@ def _construct_gateway(self, normalized_listeners: List[GatewayListener]):
...
-            )
+            listener_name = f"{norm_listener.protocol.lower()}-{norm_listener.port}"
 
-            # Add TLS config if this is an HTTPS listener
-            if norm_listener["gateway_protocol"] == "HTTPS" and norm_listener["tls_secret_name"]:
-                listener.tls = GatewayTLSConfig(
-                    certificateRefs=[SecretObjectReference(name=norm_listener["tls_secret_name"])]
-                )
+            # replace old placeholder values
+            norm_listener.name = listener_name
```
adhityaravi: IMO, its bad practice to mutate input arguments in place. In general, we cannot get an input arg, use parts of it, then mutate it by changing other properties of it. New object is the better approach here.

src/charm.py
```diff
@@ -1398,31 +1380,9 @@ def _construct_httproutes(self, http_routes: List[HTTPRoute]) -> List:
...
-                        HTTPRouteRule(
-                            matches=route["matches"],  # Already charm HTTPRouteMatch models
-                            backendRefs=route["backend_refs"],  # Already charm BackendRef models
-                            filters=route["filters"] if route["filters"] else None,
-                        )
-                    ],
-                ),
-            )
+            listener_name = f"{route.listener_protocol.lower()}-{route.listener_port}"
+            http_route_resource = route.resource
```
adhityaravi: Same comment as above. I don't necessarily think cleaner code means less code. wdyt?

src/charm.py
```diff
@@ -1450,31 +1410,11 @@ def _construct_grpcroutes(self, grpc_routes: List[GRPCRoute]) -> List:
...
-                    rules=[
-                        GRPCRouteRule(
-                            matches=route["matches"],  # Already charm GRPCRouteMatch models
-                            backendRefs=route["backend_refs"],  # Already charm BackendRef models
-                            filters=route["filters"] if route["filters"] else None,
-                        )
-                    ],
-                ),
-            )
+            grpc_route_resource = route.resource
```
adhityaravi: Same pattern as above 2 comments.

src/utils.py
adhityaravi, line 1: The changes here, I assume, are majorly done to accomodate the in-place mutating pattern in `charms.py` so I'll wait for the reply for the above comments before going into this

src/charm.py
gycsaba96: Done.

src/charm.py
gycsaba96: Done.

src/charm.py
gycsaba96: Done.

src/utils.py
gycsaba96: I fixed the in-place rewrites. The only change we could consider, is injecting all the information into the object right during creation. Then we wouldn't have rewrite vs no-rewrite problems at all.

adhityaravi (commented):
I am not entirely convinced that this refactor is adding value. 

The existing pattern is to collect the ingress information from the IPA interface and the istio ingress route interface. these both has different shapes. So, they were aggregated into a common shape using the internal utility classes. This was done so they can be deduped and conflict resolved together. Finally the aggregated, deduped, and conflict resolved listeners and routes were converted to K8s models and submitted.

the approach introduced by this PR drops some of the intermediate classes, and changes the blueprint of the other intermediate classes, this is making the code hard to read in some places for example where the deep_copy is introduced. If i, for example, compare the previous `construct_httproute` fn vs the one in this PR, the current approach makes it hard for me to understand what information does the http route resource is constructed with.

I am happy to be convinced otherwise. I think we can improve charm code for sure, but i don't think the current approach is it.

In anycase, i think a second opinion from @dstathis or @sed-i would be nice as well

gycsaba96:
> I am not entirely convinced that this refactor is adding value.

I also highlighted my concerns in the description. If we are in doubt (and it seems we are), I think we should just not merge it and close the related issue.

adhityaravi:
@gycsaba96 I have an idea - 

The current problem is that we are using the intermediary classes to basically to dedup, figure out conflicts, and then we are creating the K8s resources. right?

What if we switched the flow? Like lets use the relation data to create the K8s resource models ie directly create the HTTPRouteResource, Listener etc etc (so this way we have no intermediary classes no HTTPRoute, no need to do deep copies etc). Basically, the normalization functions can

1. take in the ingress information from both the interfaces
2. create the httproute, grpcroute, and listener resource with the k8s models
3. then dedup the listeners
4. find the conflicts between httproutes

and return them?

This brings the resource creation step forward but eliminates the intermediary classes better giving us a cleaner architecture?

WDYT? If this is what you meant by "The only change we could consider, is injecting all the information into the object right during creation. Then we wouldn't have rewrite vs no-rewrite problems at all.", I think its a great idea!

gycsaba96:
Hi @adhityaravi,

> What if we switched the flow? Like lets use the relation data to create the K8s resource models ie directly create the HTTPRouteResource, Listener etc etc (so this way we have no intermediary classes no HTTPRoute, no need to do deep copies etc). Basically, the normalization functions can [...]

This works for Listener but not for the other two classes because they need more information during deduplication. For example, `source_app` and `soure_relation` are not stored in the `HTTPRouteResource` and `GRPCRouteResource` objects, but they are essential to the deduplication process. Because of this, `HTTPRoute` and `GRPCRoute` can not be replaced with their corresponding k8s resource class.

> WDYT? If this is what you meant by "The only change we could consider, is injecting all the information into the object right during creation. Then we wouldn't have rewrite vs no-rewrite problems at all.", I think its a great idea!

What I mean is that the `utils.normalize_*` functions could already populate all the information that is later overwritten at the end. For example, Listener.allowed_routes can be created with the correct value [here](https://github.com/canonical/istio-ingress-k8s-operator/blob/1d5429b0724280b3bc748cbe22c3b684dab666a2/src/utils.py#L148) and [here](https://github.com/canonical/istio-ingress-k8s-operator/blob/1d5429b0724280b3bc748cbe22c3b684dab666a2/src/utils.py#L190), instead of filling them in at [a later stage](https://github.com/canonical/istio-ingress-k8s-operator/blob/1d5429b0724280b3bc748cbe22c3b684dab666a2/src/charm.py#L640). However, this would still not eliminate `HTTPRoute` and `GRPCRoute` because they need some additional info for the deduplication.

adhityaravi:
@gycsaba96 

The `source_app` and `source_relation` can go into the resource as custom metadata labels. That will help us not only figure out conflict but also is useful in debugging actual deployments. You can open a HTTPRoute resource and see which app and which relation it came from. Something like

```yaml
metadata=Metadata(
    name=route_name,
    namespace=namespace,
    labels={
        "istio-ingress.juju.is/source-app": app_name,
        "istio-ingress.juju.is/source-relation": relation_name,
    },
)
```

dstathis (commented):
I think I am leaning toward not keeping this as well. While it is a shame to lose work that we spend time on, I think we aren't achieving what we set out to do here. We were supposed to get rid of the intermediate classes and it seems like we can't. :(

So in light of that, we should probably just close this and move on to other things.

